PR #8, opened 2023-08-29
i18n: Add translation of zh-cn and zh-hk

naranyinyun:
Add translation of zh-cn and zh-hk


imfing:
@naranyinyun thank you for the contribution! though we still need to work on translating the exampleSite to other languages.

naranyinyun:
> @naranyinyun thank you for the contribution! though we still need to work on translating the exampleSite to other languages.

Ok，I will translate exampleSite to Chinese

PR #63, opened 2023-09-16
Add native image processing to cards

remysheppard:
I modified the `cards` shortcode to also natively handle locally stored image assets and process them. 

Ripped a little code from @bep on the [Hugo Forums](https://discourse.gohugo.io/t/0-90-0-how-to-check-for-local-resource-existence/35954) to keep from spitting out an error. 

The shortcode should still be able to handle images in the `/static/` folder as well, passing them through without processing.


layouts/shortcodes/card.html
```diff
@@ -5,6 +5,13 @@
 {{- $image := .Get "image" }}
 {{- $context := . -}}
 
+{{/*- Adding asset support for images here, so that Hugo can do its image processing magic. -*/}}
+{{- if not (urls.Parse $image).Scheme -}}
+  {{- with resources.Get $image -}}
+    {{- $image = (.Resize "800x webp q80").RelPermalink -}}
```
imfing, line 11: Can we make the resize target `800x`, format `webp` and quality `q80` configurable?

layouts/shortcodes/card.html
```diff
@@ -5,6 +5,13 @@
 {{- $image := .Get "image" }}
 {{- $context := . -}}
 
+{{/*- Adding asset support for images here, so that Hugo can do its image processing magic. -*/}}
+{{- if not (urls.Parse $image).Scheme -}}
+  {{- with resources.Get $image -}}
```
imfing, line 10: does it work with the case when users need to use images under `/static/` directory?

imfing (changes requested):

remysheppard:
Hextra can now handle local images in either the `/static/` or `/assets/` directory.

I also piped `markdownify` into subtitles and expanded the documentation on the page a bit. I'm not used to writing documentation, so if that needs to be re-written in some way, let me know.

exampleSite/assets/images/astronaut.jpg
imfing, line 1: @remysheppard Can I confirm what's the license of this image?

exampleSite/static/images/card-image-unprocessed.jpg
imfing, line 1: Is this the source of the image?
https://wallpapers.com/picture/vibey-1920-x-1080-picture-mbsgn6c608mslp91.html

imfing:
> I also piped `markdownify` into subtitles and expanded the documentation on the page a bit. I'm not used to writing documentation, so if that needs to be re-written in some way, let me know.

Thanks, looks good overall. I'll update the docs part before merging. I just have two questions about the two images used above.

remysheppard:
The images have been swapped out with the following images in the public domain, courtesy of Wikimedia Commons:

* https://commons.wikimedia.org/wiki/File:MtFuji_FujiCity.jpg
* https://commons.wikimedia.org/wiki/File:Michael_Gernhardt_in_space_during_STS-69_in_1995.jpg

exampleSite/assets/images/astronaut.jpg
imfing, line 1: Can we use a smaller sized image? We could just use the images from Unsplash, and keep the aspect ratio.
e.g. https://source.unsplash.com/featured/1200x900?landscape

otherwise the cards would look like:

image

imfing (changes requested):

remysheppard:
I adjusted the image height to be processed via crop into 1200x900. We can't remote link images because then Hugo can't process them.

imfing:
> I adjusted the image height to be processed via crop into 1200x900. We can't remote link images because then Hugo can't process them.

Apologies for the confusion earlier. The astronaut image is too big at 5Mb. We can simplify by directly using landscape images from Unsplash. Just download from this link: https://source.unsplash.com/featured/1200x900?landscape. This way, we won't need to manually adjust the aspect ratio

remysheppard:
You mean you don't want a 5mb image in your repo?

exampleSite/hugo.yaml
```diff
@@ -86,6 +86,7 @@ menu:
 
 params:
   description: Modern, responsive, batteries-included Hugo theme for creating beautiful static websites.
+  fullWidth: false
```
imfing, line 89: can you revert this

imfing (changes requested):

imfing:
> You mean you don't want a 5mb image in your repo?

No, 5Mb is fairly large in size for a binary image in Git repo. 

There's no need to use such a large image just for a thumbnail in a card component, especially when we have good alternatives like the Unsplash images that are under 200kb.

exampleSite/hugo.yaml
remysheppard: Oh sorry! I didn't mean to include that. I was messing around with the full-width issue.

imfing: also i saw there are some empty line changes, can you revert those as well?

remysheppard:
Okay sorry I understand now. I replaced both images with an unsplash image. Each one should now be under 200k

remysheppard:
Just as a note: While it is probably good practice to make sure you capitalize the command in `method` for images, it is not required. I coded that bit pretty defensively, and Hugo should automatically sanitize whatever is typed. So even if you use `method="resize"`, Hugo will convert it to `Resize` before processing.

imfing (approved):

imfing:
> Just as a note: While it is probably good practice to make sure you capitalize the command in `method` for images, it is not required. I coded that bit pretty defensively, and Hugo should automatically sanitize whatever is typed. So even if you use `method="resize"`, Hugo will convert it to `Resize` before processing.

gotcha. I used the capitalized `method` just for consistency with the Hugo built-in method names.

imfing:
Thank you for the PR! @remysheppard 🎉

Issue #51, opened 2023-09-13
Feature Request: Full and Wide Screen Mode


imfing:
### Discussed in https://github.com/imfing/hextra/discussions/45

<div type='discussions-op-text'>

<sup>Originally posted by **saurabh-mish** September 12, 2023</sup>
A configuration option for full screen mode. It will use the complete width of the screen like [this theme](https://lotusdocs.dev/docs/overview/) for instance.</div>

PR #69, opened 2023-09-20
feat: support wide and full page modes

imfing:
closes: #51 

Default:
default

Wide:
wide

Full:
full


saurabh-mish:
Thank you so much !!

PR #71, opened 2023-09-21
fix: htmlUnescape page description

iju707:
If contents has escape word(like >), description in blog list page show escape for html (like &gt;).
This often occurs on pages that contain source code.

so, add "htmlUnescape" funtion in page-description.html.


imfing:
@iju707 Thank you for the contribution!

PR #73, opened 2023-09-21
ci: upgrade `actions/checkout` from `v3` to `v4`

hugo-sid:
https://github.com/actions/checkout#checkout-v4


imfing:
@hugo-sid Thank you for the PR 😄

PR #87, opened 2023-09-24
fix: reverse order in blog list page

iju707:
Most users will want to list the most recent posts on their blog.

so, support reverse order using param `params.blog.order: desc`


imfing:
Actually, the ordering of the post is a bug, I wasn't intended to sort the blog in ascending order. Let's just default to descending order.

imfing (approved):

PR #105, opened 2023-09-27
feat: Back To Top

remysheppard:
Back to top function function mirroring Nextra's

I'm a little happy with the way Nextra implements this, as they use `opacity` for hiding and showing the button. This means that the back to top button is actually *always* clickable (seriously, go to Nextra and check). But after playing around I see why they made the concession: You have to use `opacity` or the button will *pop* into existence. There is no other way to get that nice fade effect, that I can see.

Anyway, works as intended.

Updated files:

- `toc.html` - Includes the actual button on the page.
- `en.yaml` - Includes i18n specification for `backToTop` so that text can be modified, if desired.
- `scripts.html` - Includes new script file `back-to-top.js` in the splice.
- `back-to-top.js` - VanillaJS functionality for the button.

The functionality is pretty simple: Just watch the viewport, and add/remove `opacity-0` as needed. Best I could tell from the Nextra source (I don't do React/TypeScript stuff), that's all they're doing.

Here's the JS:

```JavaScript
const backToTop = document.querySelector("#backToTop");

document.addEventListener("scroll", (event) => {
  if (window.scrollY > 300) {
    backToTop.classList.remove("opacity-0");
  } else {
    backToTop.classList.add("opacity-0");
  }
});


function scrollUp() {
  window.scroll({
    top: 0, 
    left: 0, 
    behavior: 'smooth'
  });
}
```

Let me know if I need to tweak anything!

### DIFF
--- a/assets/js/back-to-top.js
+++ b/assets/js/back-to-top.js
@@ -1,0 +1,18 @@
+const backToTop = document.querySelector("#backToTop");
+
+document.addEventListener("scroll", (event) => {
+  if (window.scrollY > 300) {
+    backToTop.classList.remove("opacity-0");
+  } else {
+    backToTop.classList.add("opacity-0");
+  }
+});
+
+
+function scrollUp() {
+  window.scroll({
+    top: 0, 
+    left: 0, 
+    behavior: 'smooth'
+  });
+}
--- a/i18n/en.yaml
+++ b/i18n/en.yaml
@@ -2,4 +2,6 @@ onThisPage: "On this page"
 editThisPage: "Edit this page on GitHub →"
 lastUpdated: "Last updated on"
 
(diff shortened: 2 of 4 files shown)


imfing:
Neat implementation! I appreciate your quick PR.

remysheppard:
> Just need to recompile the css

I'm sorry, is this something I need to do? And if so, how? I thought the CSS compiled on build

imfing:
https://github.com/imfing/hextra/blob/main/.github/CONTRIBUTING.md#compile-the-styles

running this command should compile the [`assets/css/compiled/main.css`](https://github.com/imfing/hextra/blob/main/assets/css/compiled/main.css)

nvm, I can merge the PR and help with that. Thanks for the contribution!

PR #130, opened 2023-10-04
feat: support empty prev/next pagination

iju707:
It's related #127

It support empty prev/next pagination.
using by front matter bellow.

```
---
prev: false
next: false
---
```

but I dont know how to solve in blog single page layout.
caused by prev and next are reversed, so many users will be  confused.


imfing:
@iju707 thank you for the PR!

PR #137, opened 2023-10-12
Add docs for using Hextra as Git submodule, refactor docs for using Hextra as Hugo module

hugo-sid:
## Issue with this PR

At present, an info callout (which is part of this PR) appears like this:

image

All my attempts to fix this went in vain. I observed that there is a leading empty `<p>` tag and trailing empty `<p>` tag within the markup of the shortcode. By markup I am referring to the HTML rendered by the shortcode.

It appears correctly when I delete these empty tags.

image

Trimming newline characters while processing the Hugo shortcode might be a fix for this.

## Enhancement

I realized that the getting-started page looks quite lengthy when the Git submodule docs is added. This can be a bad user experience especially for beginner/new users of Hextra. For a better user experience, I suggest that we split 'Setup Hextra as Hugo module' section and 'Setup Hextra as Git submodule' section to new files (under getting started).

Then we can link these pages using cards on the getting-started page.

Let me know your thoughts @imfing. I can amend the PR accordingly.

### DIFF
--- a/exampleSite/content/docs/getting-started.md
+++ b/exampleSite/content/docs/getting-started.md
@@ -74,7 +74,7 @@ $ hugo mod init github.com/username/my-site
 $ hugo mod get github.com/imfing/hextra
 ```
 
-Configure your `hugo.yaml` file to use Hextra theme by including the following:
+Configure `hugo.yaml` to use Hextra theme by adding the following:
 
 ```yaml
 module:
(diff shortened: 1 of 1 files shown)


imfing:
Thank you for the PR! I'll take a look tomorrow.

For the displaying issue you mentioned, I think let's avoid using callout for that part at the moment. 
We should address it in a separate issue and PR.

hugo-sid:
> Thank you for the PR!

You're welcome! 

> For the displaying issue you mentioned, I think let's avoid using callout for that part at the moment.

Done.

> i've create https://github.com/imfing/hextra/issues/140 for the callout display issue


> and it's addressed in https://github.com/imfing/hextra/pull/141

That's great!

PR #142, opened 2023-10-13
fix: add missing translation in blog list

iju707:
In blog list page,
"Read More →" is not a variable that can be translated.


iju707:
I found that missed variables in en.yaml

imfing:
thank you, I'll take a look in a few days

imfing (approved):

PR #153, opened 2023-10-26
feat: Improve Cards Link Behavior

illlustr:
currently card will "refresh" the page if link is not provided.
this PR addressing the issue and add similar behavior to `feature-card`

changes:
- unclickable `card` when no link is provided.
- clickable `feature-card` when a link is provided.


imfing:
looks good, thanks! i've updated the `feature-card` to just use `<a>` tag

PR #155, opened 2023-10-27
fix(style): remove sidebar item extra padding and gap

imfing:
* remove the `pt-1` from the `div` itself
* apply the `pt-1` when the `div` container is open
* remove the `gap-1` from the `li` item as we set it in the `ul` already


imfing:
Before:

image


After:

image

PR #166, opened 2023-10-29
chore(i18n): add French translation

c2tz:
I didn't know if I should use en.yml or zh-cn.yml, so I took the one that was the most complete. And what is the `showcase:`  value ?


imfing:
@c2tz thanks for the PR! 

the "showcase" entry was for the top navigation bar:

c2tz:
Yeah I just saw it while browsing 😄 .  I think my translation is pretty good, at worst it can be improved afterwards.

c2tz:
@imfing that ok for me

i18n/fr.yaml
```diff
@@ -0,0 +1,20 @@
...
+dark: "Sombre"
+documentation: "Documentation"
+editThisPage: "Modifier cette page sur GitHub →"
+hugoDocs: "Docs Hugo ↗"
+lastUpdated: "Dernière modification"
+light: "Clair"
+more: "Plus"
+noResultsFound: "Pas de résulats trouvés"
+onThisPage: "Sur cette page"
+onThisPage: "Sur cette page"
```
imfing, line 16:
```suggestion
onThisPage: "Sur cette page"
```
duplicated

i18n/fr.yaml
```diff
@@ -0,0 +1,20 @@
...
+changeTheme: "Thème d'affichage"
+copyright: "© 2023 Hextra Project."
+dark: "Sombre"
+documentation: "Documentation"
+editThisPage: "Modifier cette page sur GitHub →"
+hugoDocs: "Docs Hugo ↗"
+lastUpdated: "Dernière modification"
+light: "Clair"
+more: "Plus"
+noResultsFound: "Pas de résulats trouvés"
```
imfing, line 14:
```suggestion
noResultsFound: "Pas de résultats trouvés"
```
typo

i18n/fr.yaml
```diff
@@ -0,0 +1,20 @@
...
+onThisPage: "Sur cette page"
+onThisPage: "Sur cette page"
+poweredBy: "Propulsé par Hextra"
+readMore: "Lire plus →"
+searchPlaceholder: "Rechercher..."
+showcase: "Showcase"
```
imfing, line 20: should it be translated to "Présentation"?

imfing (changes requested):

i18n/fr.yaml
c2tz: I don't think so, because Présentation is too vague, it could be anything, in French we use showcase, the meaning is not unknown, or Featured, but there's no exact translation.

imfing: got it, thanks for explaining! merci beaucoup :)

imfing:
@c2tz Merci beaucoup!

PR #172, opened 2023-10-30
i18n: Add language strings for Norwegian bokmål and Norwegian nynorsk

bep:
The two main written languages in Norway.


imfing:
@bep thank you so much ❤️

PR #180, opened 2023-11-01
i18n: update es.yaml

tehuel:
Update spanish translations, using keys from `zh-cn.yaml`.


imfing:
@tehuel thank you for the contribution!

PR #184, opened 2023-11-02
Create Readfile Shortcode

juliamrch:
## What is a Readfile shortcode

Readfile shortcodes are useful to [re-use content across a Hugo site](https://gohugo.io/templates/files/#use-readfile). In[ Clever Cloud](https://www.clever-cloud.com) documentation (which is being revamped using this template), we use them heavily. 

I've therefore created a Readfile shortcode and thought you could find it useful integrated in Hextra theme.

## Bonus

A dedicated page and example in your documentation site has been added as well.

### DIFF
--- a/exampleSite/content/docs/guide/shortcodes/readfile.md
+++ b/exampleSite/content/docs/guide/shortcodes/readfile.md
@@ -1,0 +1,19 @@
+---
+title: Readfile
+---
+
+Readfiles are useful for content templating. I allows you to re-use a file's content in another file.
+
+## Example
+
+{{< readfile file="readfile-example.md" >}}
+
+## Usage
+
+Put your Markdown templates in `/static/templates/`.
+
+Print the content in another file with the following shortcode:
+
+```
+{{</* readfile file="your-file.md" */>}}
+```


imfing:
@juliamrch Hi Julia, thank you! 
I would love to include that in the theme. Let me know when you complete in this PR. 

here's an example you could use to implement that: https://github.com/google/docsy/blob/main/layouts/shortcodes/readfile.html

bep:
My 50 cents about this; including files in Markdown using the `readFile` func is not a great idea, and that will be even more true in the next version of Hugo. I would recommend using a variation of 


```handlebars
{{ $page := site.GetPage $path }}
{{ $page.RenderShortcodes }}
```

If you then use the shortcode using the `{{%` delimiter, the included content will be a part of the outer document, which is important re. ToC, footnotes etc.

I would even use a variation of the above (using `.Content`) to include non-content files, the main reason being that Hugo can track its usage and be smart about partial rebuilds.

juliamrch:
> My 50 cents about this; including files in Markdown using the `readFile` func is not a great idea, and that will be even more true in the next version of Hugo. I would recommend using a variation of
> 
> ```
> {{ $page := site.GetPage $path }}
> {{ $page.RenderShortcodes }}
> ```
> 
> If you then use the shortcode using the `{{%` delimiter, the included content will be a part of the outer document, which is important re. ToC, footnotes etc.
> 
> I would even use a variation of the above (using `.Content`) to include non-content files, the main reason being that Hugo can track its usage and be smart about partial rebuilds.

Why do you think it would be a bad idea?

imfing:
Implemented in #198

PR #189, opened 2023-11-05
Fix showcase starter template URL

saurabh-mish:
Current url returns a 404. Updated URL goes to template homepage.


imfing:
@saurabh-mish thank you for the PR!

PR #212, opened 2023-11-17
chore(style): better footer logo alignment

c2tz:
I just added a -2.5px value for the footer logo because the base logo tended to be a little too low and not aligned with the "Hextra". Not sure it's very useful, but here it is anyway. On the left, the modified version, and on the right, the basic version.

chrome_myg9wuGcJo


imfing (changes requested):
Thanks for the PR! It looks like you're editing the auto-compiled CSS. For styling changes, please use [Tailwind CSS classes](https://tailwindcss.com/docs/utility-first) as outlined in our [Contributing Guide](https://github.com/imfing/hextra/blob/main/.github/CONTRIBUTING.md#compile-the-styles).

c2tz:
ok sorry it's better ?

c2tz:
wait i enlarged too much the svg icon i thought i did well because it would have lost me an alignment but it overflows because it is more than 14px

c2tz:
ok for me 

chrome_9iX0q7Gtzc

imfing:
@c2tz thanks, that should be it, the change looks good to me 👍

imfing (approved):

PR #215, opened 2023-11-21
docs: fix the documentation error about editURL

longlin10086:
The `.Params.editURL` can't find the value when using 
```
---
params:
  editURL: "https://"
---
```
in `front matter`. 
It works correctly with
```
---
editURL: "https://"
---
```


imfing (approved):

imfing:
@longlin10086 thank you for the fix

PR #219, opened 2023-11-23
feat: add ability to hide posts in blog listing

otherpaco:
I wanted to use the starting page as blog.
But then all entries like "/about" or "/imprint" will show up as posts as they all are in the same folder `content`.

I added a param `hideOnBlog` to the front matter. That parameter will prevent the file from being listed in the blog overview but it is still accessible.

### DIFF
--- a/exampleSite/content/blog/_index.md
+++ b/exampleSite/content/blog/_index.md
@@ -8,3 +8,5 @@ title: "Blog"
   {{< icon name="rss" attributes="height=14" >}}
 {{< /hextra/hero-badge >}}
 </div>
+
+Checkout the unlisted post, find it here `/blog/hidden`.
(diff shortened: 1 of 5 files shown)


imfing:
Thank you, I like this idea. I will take a look

otherpaco:
Dear Xin, it realized there is much more to do and check to use the starting page as blog.
I will keep on modifying my blog an will come back with a better PR or a feature request.

I removed the new parameter because I can simply use

`{{- if eq .Params.type "blog" -}}`

instead of

`{{- if not .Params.hideOnBlog -}}`

But there are more places I would need to add that e.g. `layouts/partials/components/pager.html`

Maybe it would be easier going with `collections.Where` [gohugo docs](https://gohugo.io/functions/collections/where/)

Feel free to close this PR and I could move the info to an issue if you prefer it that way.

Have a nice weekend

PR #221, opened 2023-11-28
Add children main menu support

fooying:
Add children menu support in main menu


hae-andrew-young:
This would be very nice to have

PR #244, opened 2024-01-05
chore: add Codeberg icon

verbumfeit:
This PR adds a Codeberg icon that can be shown in the navbar, just like the GitHub icon.


imfing:
@verbumfeit thank you so much!

PR #245, opened 2024-01-07
fix(docs): add editURL.enable = true

ytjohn:
The MR documents the requirement to enable editURL setting.

[A  prior commit](https://github.com/imfing/hextra/commit/04a5c7378a867111e91e3cd6425ed132889a5ad1#diff-5311bf4e1519696010f8298385d9de30210dfcd9dee8785f6deba3a084b62eecL26) the behavior changed from a default of "disable = false" to a default of "enabled = false".


imfing:
@ytjohn thank you for catching this!

PR #246, opened 2024-01-09
chore(i18n): add Czech translation

pavel-francirek:
Czech translation


imfing:
@pavel-francirek thank you!

PR #258, opened 2024-01-17
Correct symbol in details shortcode

thisven:
Hi,

many thanks for creating this Hugo theme. :smiley:

While getting to know it's various features, I believe there's a typo in the [documentation of the details shortcode](https://imfing.github.io/hextra/docs/guide/shortcodes/details/).

### DIFF
--- a/exampleSite/content/docs/guide/shortcodes/details.md
+++ b/exampleSite/content/docs/guide/shortcodes/details.md
@@ -8,36 +8,36 @@ A built-in component to display a collapsible content.
 
 ## Example
 
-{{% details title="Details" %}}
+{{< details title="Details" >}}
 
 This is the content of the details.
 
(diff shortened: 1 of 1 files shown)


imfing:
@thisven actually it's not typos. See Hugo's documentation:

> Shortcodes using the % as the outer-most delimiter will be fully rendered when sent to the content renderer. This means that the rendered output from a shortcode can be part of the page’s table of contents, footnotes, etc.

https://gohugo.io/content-management/shortcodes/#shortcodes-with-markdown

the rendering was explicitly executed in `details` shortcode:

https://github.com/imfing/hextra/blob/09e9f52019edbfece778db2a493088e535d7676c/layouts/shortcodes/details.html#L9

thus, both `<` and `%` would work here.

PR #262, opened 2024-01-19
showcase: clever cloud documentation

juliamrch:
Added [Clever Cloud Documentation website](https://developers.clever-cloud.com) in the showcase page of Hextra site, as suggested [here](https://github.com/imfing/hextra/discussions/259#discussioncomment-8171964). 

Image is hosted in our storage object, to avoid weighing Hextra's site with an additional PNG, but we can relocate it in `/static` if it suits you better.


imfing:
@juliamrch Thank you, Julia! 🙌 

> Image is hosted in our storage object, to avoid weighing Hextra's site with an additional PNG, but we can relocate it in /static if it suits you better.

It's fine. For the other images in the showcase, I uploaded them on GitHub's Issue, hence they are served by GitHub's CDN.

Issue #275, opened 2024-02-03
Allow disable generating full mobile navigation tree


imfing:
### Discussed in https://github.com/imfing/hextra/discussions/274

<div type='discussions-op-text'>

<sup>Originally posted by **dogweather** February  3, 2024</sup>
I'm trying to switch from Ananke Theme to Hextra, but I have a weird issue. Every page has a hidden list of links to all the other pages on the site. I have 20,000 pages, so this is a big problem. 

It's not just simple `<a>` tags either: here's a screenshot from the middle of a simple About page. As far as I can tell, I haven't enabled any strange configurations. And what are these line number spans?

<img width="879" alt="Screenshot 2024-02-02 at 11 54 55 PM" src="https://github.com/imfing/hextra/assets/150670/785204ab-6839-41f6-b4ad-45d4d1d214d4">

The page itself doesn't show anything strange. A megabyte of HTML is hidden in there, though:

<img width="1079" alt="Screenshot 2024-02-02 at 11 58 16 PM" src="https://github.com/imfing/hextra/assets/150670/aa040f95-00fa-441e-a284-8e61271e5ac8">

</div>

PR #287, opened 2024-02-12
feat: add parameter to empty sidebar on mobile

davlgd:
Fix #274 #275

- Add `sidebar.emptyMobile` in `site.Params`
- Set it to `false` in the example site
- Add the a condition in the sidebar layout
- If the parameter is not set or false, the sidebar is shown

I check on the example site, work as expected


davlgd:
@imfing is it still ongoing or should I close ?

PR #290, opened 2024-02-13
feat: add configurable edit link position

### LINKED ISSUE #241: Make "Edit this page" button always visible
**Feature Description**

Ability to choose the location of the "Edit this page" link.

**Problem/Solution**

Currently, below the table of contents in the right side bar, an "Edit this page" button can be displayed. However, this button is not visible if the window is too narrow to display the TOC. Especially on mobile, there is no way to access the "Edit this page" button.

If the location of this button could be freely chosen (e.g. between right side bar, left side bar, navigation bar, bottom of page, ...) it could be configured such that it is always visible/can always be accessed.

**Alternatives Considered**

Building a custom shortcode to include this link. This would limit its location to the main page content.

**Additional Context**

Related discussion: #48

davlgd:
Fix #241 (I followed instructions seen in this conversation) 

- Add a `where` parameter to `editURL`
- Allow user to choose between `toc` (default) and `single`
- Add to single through last-updated partial


exampleSite/hugo.yaml
```diff
@@ -133,6 +133,7 @@ params:
 
   editURL:
     enable: true
+    where: "toc" # single | toc
```
imfing, line 136: A nitpick, but I'd prefer a more descriptive name for this field, such as `location` or `placement`.

layouts/partials/components/last-updated.html
```diff
@@ -1,9 +1,27 @@
 {{- $lastUpdated := (T "lastUpdated") | default "Last updated on" -}}
+{{- $editThisPage := (T "editThisPage") | default "Edit this page"}}
 
 {{- if site.Params.displayUpdatedDate -}}
+  <div class="mt-12 mb-8 block text-xs text-gray-500 ltr:text-right rtl:text-left dark:text-gray-400">
+  {{- if and (site.Params.editURL.enable) (eq site.Params.editURL.where "single") -}}
```
imfing, line 6: I think we should extract this edit url piece into a partial component, since it's been used both here and in TOC template.

imfing (dismissed):
Thank you for the PR! 

In addition to the points mentioned earlier, it would be greatly beneficial to include documentation for this new config option. 

I'm also open to finish the development of this feature. While it may take a bit longer, I plan to add it into the upcoming minor release.

imfing (changes requested):
Per above

layouts/partials/components/edit-url.html
```diff
@@ -0,0 +1,19 @@
+{{- $editThisPage := (T "editThisPage") | default "Edit this page"}}
+
+<div class="mt-12 mb-8 block text-xs text-gray-500 ltr:text-right rtl:text-left dark:text-gray-400">
```
imfing, line 3: can you double check the classes are consistent with what we had previously in the `toc.html`?

imfing (changes requested):
the styling is broken:
image

could you update the PR to fix it?

layouts/partials/components/edit-url.html
davlgd: It should be fine now (and rebased)

layouts/partials/toc.html
```diff
@@ -23,23 +23,8 @@
...
-              {{/* `.FileInfo.Meta.SourceRoot` is a Hugo internal field, e.g. `/path/to/repo/content/en/` */}}
-              {{- $sourceDir := replace (strings.TrimPrefix .FileInfo.Meta.BaseDir .FileInfo.Meta.SourceRoot) "\\" "/" -}}
-              {{- $sourceDir = strings.TrimPrefix "/content" $sourceDir -}}
-              {{- $path := replace .Path "\\" "/" -}}
-              {{- $editURL = urls.JoinPath $editURL $sourceDir $path -}}
-            {{- end -}}
-          {{- end -}}
-          <a class="hx-text-xs hx-font-medium hx-text-gray-500 hover:hx-text-gray-900 dark:hx-text-gray-400 dark:hover:hx-text-gray-100 contrast-more:hx-text-gray-800 contrast-more:dark:hx-text-gray-50" href="{{ $editURL }}" target="_blank" rel="noreferer">{{ $editThisPage }}</a>
-        {{- end -}}
+      <div class="{{ $borderClass }} sticky bottom-0 flex flex-col items-start gap-2 pb-8 dark:border-neutral-800 contrast-more:border-t contrast-more:border-neutral-400 contrast-more:shadow-none contrast-more:dark:border-neutral-400">
```
imfing, line 26: the classes here also need to be prefixed with `hx-`

imfing (changes requested):

layouts/partials/components/edit-url.html
```diff
@@ -0,0 +1,21 @@
+{{- $editThisPage := (T "editThisPage") | default "Edit this page"}}
+{{- $linkClass := "mb-4 block text-xs text-gray-500 ltr:text-right rtl:text-left dark:text-gray-400" -}}
+{{- if and (site.Params.editURL.enable) (eq site.Params.editURL.location .source) -}}
+  {{- $editURL := site.Params.editURL.base | default "" -}}
+  {{- with .Params.editURL -}}
```
imfing, line 5: the `.` here will not be able to access the page object. we should pass it down from where it's called. you can refer to other partials for how this can be achieved.

imfing (changes requested):

layouts/partials/components/edit-url.html
imfing: as a result, the edit url will not be constructed correctly. for example: https://deploy-preview-290--hugo-hextra.netlify.app/docs/

imfing:
did you update the partial?

davlgd:
Sorry I misunderstood your comment, it should be fixed now

imfing:
> Sorry I misunderstood your comment, it should be fixed now

thanks will take a look

imfing (changes requested):

davlgd:
@imfing is it still ongoing or should I close ?

PR #310, opened 2024-02-27
feat: add sectionsTitles parameter

davlgd:
If the website is composed of multiple sections, we could want to define different titles templates for each of them. For example: 

- documentation: Title - My documentation 
- blog: Title - My blog
- guides: Title - My guides

With this PR, the user can define a new entry `params`, `sectionsTitles` to define per section title, used in the title of section pages. If none is defined, the default title is used.

Example: 

```
title: My documentation

params:
  sectionTitles: 
    blog: My blog
    guides: My guides
```


davlgd:
> thanks! we need to take into account that our site supports multiple languages. the current approach might not work for multi-language sites. we should probably figure out a different approach.

For multi-languages it works too, you just have to put `sectionTitles` in a `languages` > `lang code` > `params` section. For example for a french version of above parameters:

```
languages:
  fr:
    params: 
      sectionTitles: 
        blog: Mon blog
        guides: Mes guides
```

davlgd:
@imfing is it still ongoing or shoud I close ?

PR #319, opened 2024-03-13
feat: support customize flexsearch tokenize option

loenvom:
The default tokenize option 'forward' for flexsearch is good, and have better performance compare with option 'full'. 
**But it can not support some partial search, so i think it may be a good idea to set up a param in `settings.yaml` for user to decide it.** 
Here are the search effect diagrams of two different configurations. The first one with option 'full' can get more results than the second one with option 'forward'.
image
image

### DIFF
--- a/assets/js/flexsearch.js
+++ b/assets/js/flexsearch.js
@@ -172,8 +172,9 @@ document.addEventListener("DOMContentLoaded", function () {
 
   // Preload the search index.
   async function preloadIndex() {
+    const tokenize = '{{- site.Params.search.flexsearch.tokenize | default  "forward" -}}';
     window.pageIndex = new FlexSearch.Document({
-      tokenize: 'forward',
+      tokenize,
       cache: 100,
       document: {
         id: 'id',
@@ -183,7 +184,7 @@ document.addEventListener("DOMContentLoaded", function () {
     });
 
     window.sectionIndex = new FlexSearch.Document({
-      tokenize: 'forward',
+      tokenize,
       cache: 100,
       document: {
         id: 'id',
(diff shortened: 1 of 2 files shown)


exampleSite/hugo.yaml
```diff
@@ -136,6 +136,9 @@ params:
     flexsearch:
       # index page by: content | summary | heading | title
       index: content
+      # full | forward | reverse | strict 
+      # to know more: https://github.com/nextapps-de/flexsearch/blob/540d25e97117c73b1ce9e50e54423e5d8140a710/README.md#tokenizer
```
imfing, line 140:
```suggestion
      # https://github.com/nextapps-de/flexsearch/#tokenizer-prefix-search
```

assets/js/flexsearch.js
```diff
@@ -172,8 +172,9 @@ document.addEventListener("DOMContentLoaded", function () {
 
   // Preload the search index.
   async function preloadIndex() {
+    const tokenize = window?.hextra?.search?.flexsearch?.tokenize ?? 'forward'
```
imfing, line 175: we could use hugo template here

for example: https://github.com/loenvom/hextra/blob/feat/custom_flex_search_tokenize/assets/js/flexsearch.js#L20

layouts/partials/scripts.html
```diff
@@ -1,3 +1,9 @@
+/* inject params to window.hextra for context in js */
+<script>
```
imfing, line 2: if we use hugo template in the `flexsearch.js`, then no need to inject this global variable here

imfing (changes requested):
Thanks, I think it's a great addition to the search feature.

We could leverage Hugo templates in the `flexsearch.js` rather than injecting a global variable.

loenvom:
I'am glad that you like it! I have already submitted a new commit, please help review it again. ✨

imfing (approved):
great! looks good to me 👍

imfing:
We also need to document this new option once it's release.

PR #320, opened 2024-03-14
Make copy button visible on code blocks

pchalamet:
Hello,

copy button is not visible on code blocks. Here is a screenshot from official documentation (https://imfing.github.io/hextra/docs/guide/syntax-highlighting/):

image

This PR makes the button visible:

image

⚠️ please note I'm not satisfied with the behavior: it looks like a check mark shall be displayed for 1s after click. This is still not the case. I've tried to hack the css and js but I must admit I do not know much about tailwind & css. Probably this worths a small investigation on your side !

Thanks.

### DIFF
--- a/layouts/_default/_markup/render-codeblock.html
+++ b/layouts/_default/_markup/render-codeblock.html
@@ -13,7 +13,7 @@
   {{- else -}}
     <pre><code>{{ .Inner }}</code></pre>
   {{- end -}}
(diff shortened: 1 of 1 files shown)


imfing:
hi, thanks for the PR. The current behavior is intended. 

Please use https://imfing.github.io/hextra/docs/advanced/customization/#custom-css to override the style based on your own need.

pchalamet:
Hello thanks for your answer. In that case it's worth changing this on docs:
image

I will close the PR tomorrow if it stays as is.

imfing:
@pchalamet thank you, I wasn't aware that this is a bug. The copy code button should be visible when the mouse is hovering over the code block. 

This issue has been addressed in #323

pchalamet:
Thanks.

PR #326, opened 2024-03-20
made it work

Visual-Dawg:


Visual-Dawg:
What are your thoughts on this?

imfing:
> It works, but I set it as a draft as this is a breaking change as with this users need to provide a whole palette to customize the primary color.

@Visual-Dawg thanks, like you said, asking users to provide a whole color palette which might be too much work. it also increases the complexity for maintaining.

i think what we can do is adding several custom utility classes in tailwind config to allow overriding at least backgrounds plus some other visual elements via CSS variables first.

Visual-Dawg:
Sorry for answering so late. 
I think waiting for Tailwind v4 might make sense, as this will ship with variables for the purpose of themeing.

Which should be much easier

PR #334, opened 2024-03-26
fix: language-switch.html creates links for missing translations

dogweather:
## In a nutshell

This changes the language switch to range over the page's `.Translations` instead of the site's `.Languages`.

## The problem

Some multilingual sites don't have translations for all pages. The language switch, however, stil lists all languages, creating broken links to these non-existent pages.

## The solution

Filter the options to only allow jumping to pages that exist.

I'm migrating my site from Ananke to Hextra. In Ananke, I wrote a custom language switcher which has this behavior:

* Most pages, for example [the Ruby Cookbook](https://forkful.ai/en/ruby/), have been translated into all 21 site languages.
* [The About page](https://forkful.ai/en/about/), however, has only been translated into 5 languages.

The UX is smooth: I think that most site visitors don't notice the change in the widget. IMO, this is preferable to showing an incorrect list of translations.

Thanks!


imfing:
Thanks, I think it makes sense. 
Originally the translations menu is to show a list of available translations for the entire site, instead of for single page. 

As this is a breaking change, I will spend some time to test a few scenarios. Also we need to update docs to reflect this change.

dogweather:
Yeah, I can imagine that the current mechanism might make sense in some scenarios? I'm not sure, to be honest, if this is a change that makes sense for everyone. (?)

PR #337, opened 2024-03-29
feat: shortcode for Jupyter Notebook

imfing:
Experimental shortcode to include jupyter notebook in a page. In the future, it will enable using a notebook file as an individual page.

Example usage:

```text
---
title: Jupyter Notebook
math: true
---

{{< jupyter "Intro.ipynb" >}}
```

image


dogweather:
Very cool, thank you.

MattDodsonEnglish:
Hi @imfing This is great! I just tested this branch for a use case and it worked perfectly. If I had any complaint, it'd be that I'd prefer host the notebooks in the same directory, like `static`, and I couldn't figure it out with page.Resources. But that's a minor thing. 

Just wondering, is there any reason you don't want to merge this to main? I would start using it right away.

MattDodsonEnglish:
> . If I had any complaint,

Ah, one other: it seems the headings don't show in the ToC, having tried with both `%` and `<` tokens. I'm not sure how to fix that one yet, but I'll dig into the rendering logic and see if I can find a solution.

imfing:
@MattDodsonEnglish Thank you for the feedback. 

> I would start using it right away.

I think I will probably merge this as is once I resolve the conflicts.

> Just wondering, is there any reason you don't want to merge this to main? 

I didn't figure out a way to make Jupyter work nicely in Hugo so that we can essentially use jupyter notebooks instead of `.md` files. But this can be addressed separately.

> If I had any complaint, it'd be that I'd prefer host the notebooks in the same directory,

This can be supported as what we can do for images: https://imfing.github.io/hextra/docs/guide/organize-files/#add-images

> Ah, one other: it seems the headings don't show in the ToC

I'll try to address this, but I'm not sure if this is a easy thing.

MattDodsonEnglish:
> > Ah, one other: it seems the headings don't show in the ToC
> 
> I'll try to address this, but I'm not sure if this is a easy thing.

From what I tried, I don't know how easy this is to do it elegantly. Using something like this with `%` works to make a ToC, but it causes a lot other issues, like requiring two lines two surround all markdown blocks, and things like embedded callout shortcodes fail.

```
      {{- if $toc -}}
        {{- $sourceContent -}}
      {{- else -}}
        {{- $sourceContent | $page.RenderString -}}
      {{- end -}}
```

MattDodsonEnglish:
One more request: would it be possible to keep the copy button in the cells? It's a nice UX boost for readers who just want to grab a little sample from the wider notebook flow.

imfing:
@MattDodsonEnglish yes, see the latest demo page: https://deploy-preview-337--hugo-hextra.netlify.app/docs/guide/shortcodes/jupyter/

imfing:
@MattDodsonEnglish let me know if things don't work as expected 🙌

PR #353, opened 2024-04-06
docs(i18n): add zh-cn translation of RSS button

kowyo:
I notice that RSS button is missed when switching language to Chinese in the documentation of Hextra. It's added now in this PR, hope it helps.


imfing (approved):
Great. Thank you!

Issue #239, opened 2024-01-02
Allow Shortcodes in Copyright Footer


StefanKoell:
**Feature Description**

It would be great if we can also use shortcodes, specifically for displaying the current year. 

**Problem/Solution**

Right now, you have to "hard code" the year (or I am not able to figure out how to make shortcodes work).

**Alternatives Considered**

As a workaround, I copied/modified the footer.html to use the shortcode to display the current year.

**Additional Context**

Since printing the current year is a common scenario, it would be nice if the theme would allow such thing out of the box or provide an option to display the current year automatically.

imfing:
Hi, I think we could put shortcodes in the copyright footer.

e.g. `layouts/shortcodes/year.html`:

```go
{{- time.Now.Year -}}
```

use this shortcode in `i18n/en.yaml`

```yaml
copyright: "© {{`{{% year %}}`}} Your Project"
```

PR #355, opened 2024-04-07
feat: add dynamic copyright year handling

kowyo:
One of our organization member finds that the copyright year of both our website and Hextra doc is still "2023" yet. Hence I make some changes to theme to fix this.

This PR implements dynamic handling of the copyright year in the footer template. It fetches the beginning year from the configuration file(config.yaml) and dynamically updates the copyright text based on the current year.

The convention is referred from [Copyrightlaws.com](https://www.copyrightlaws.com/copyright-symbol-notice-year/#:~:text=The%20general%20rule%20is%20that,to%20the%20public%20without%20restriction.). Where it mentions that:

> For constantly evolving websites and blogs that contain works published over several years, the notice may include a range of years (e.g., 2011–2023), starting from the date of the oldest published elements and ending with the date of the newest published elements.

Additionally, I also modify the exampleSite as well as i18n files to adapt to changes of this new feature. Let me know if I need more work on it. All code has been tested carefully.


imfing:
Hi, thank you for the PR. I guess https://github.com/imfing/hextra/issues/239 is a relevant issue.

I noticed that in the footer template
https://github.com/imfing/hextra/blob/e4cdabff12bb516b0ba977968da9d2d2ed85b8bc/layouts/partials/footer.html#L32

```
{{- $copyright := (T "copyright") | default "© 2023 Hextra." -}}
{{ $copyright | markdownify }}
```

the `markdownify`[^1] will render the `copyright` string as markdown content which also renders the shortcode if any.

That means if we have a shortcode, e.g. `layouts/shortcodes/year.html` that gives us the current year:

```go
{{- time.Now.Year -}}
```

we could use this shortcode in `i18n/en.yaml`

```yaml
copyright: "© {{`{{% year %}}`}} Your Project"
```

similarly we can have a `since` shortcode, and change the `copyright` field to:

```yaml
copyright: "© {{`{{% since %}}`}}-{{`{{% year %}}`}} Your Project"
```

https://gohugo.io/functions/time/ has docs for the time functions

If this works for you, we could use this PR to add better docs for this, and close https://github.com/imfing/hextra/issues/239

---

[^1]: https://gohugo.io/functions/transform/markdownify/

kowyo:
Thank you for your advice! It is a great idea to utilize Hugo's features to implement this function. However, it may seem a little complex to add a specific shortcode to implement a simple function like this one. Actually I also referred [another hugo theme](https://github.com/CaiJimmy/hugo-theme-stack/blob/master/layouts/partials/footer/footer.html), which inplement this feature in a more concise and standard manner.

I apologize if my code is not standard enough cause I have little formal opportunities to learn frontended development knowledge in school. Therefore, I suppose you could leave this pull request open for a more refined implementation. Nonetheless, I appreciate your thorough review.

imfing:
> However, it may seem a little complex to add a specific shortcode to implement a simple function like this one. Actually I also referred [another hugo theme](https://github.com/CaiJimmy/hugo-theme-stack/blob/master/layouts/partials/footer/footer.html?rgh-link-date=2024-04-07T16%3A26%3A15Z), which implement this feature in a more concise and standard manner.

My intention was to let users easily tweak the copyright notice to their liking, without being stuck to the predefined `© startYear-currentYear SOME TEXT`. So, if someone wants to go with “All rights reserved. Copyright Company Foo,” they totally can.

Regarding the standard format, I'm still unsure about it. First, users are able to hardcode the years under `i18n/en.yaml` like this:

```yaml
copyright: "© 2023-2024 YOUR PROJECT"
```

This way, they don't have to change the start year, and updating the current year is just a once-a-year thing.

And if users are asking dynamically generated current year, they can do so via the the utility `year` shortcode I mentioned earlier:

```yaml
copyright: "© 2023-{{`{{% year %}}`}} YOUR PROJECT"
```

From my point of view, this is slightly more easier than modifying the `params.footer.since` plus `i18n/en.yaml` just to update the copyright footer.

> I apologize if my code is not standard enough

no worries at all. this is a great discussion!

kowyo:
Now I understand you. It does make it easier for user if they could simply edit in i18n.yaml. I will try to implement this in the coming days.

jonasreiher:
I like this idea, I'm not sure what the best solution is, though (I'm also not a Hugo pro). From a user perspective, it would probably be most intuitive to only edit `i18n/en.yaml` and have something without extra shortcodes, like
```yaml
copyright: "© 2023–{{ time.Now.Year }} Your Project"
```

imfing:
> have something without extra shortcodes

unfortunately afaik it's not possible to use hugo template functions in Markdown content

Issue #231, opened 2023-12-14
Giscus widget refused  to connect. May be a bug of the method hextra deal with i18n folder.


librehugohu:
You will get more basic info in issue #228. Today I found another bug about i18n folder and giscus.

As #228 showed, I use i18n folder to manage i18n markdown files, now every files is under `./content/zh-CN` folder(I haven't write texts in English or French, but I want to setup i18n folder structure firstly). Besides I configured giscus properly, but the giscus widget can't be loaded. 

![image](https://github.com/imfing/hextra/assets/51823222/c968d382-c4b8-410f-80eb-16b8c5a71f98)

After check Network tab of Dev tools of Chrome, we can see this result.

![image](https://github.com/imfing/hextra/assets/51823222/903de97a-42a7-4835-9bf6-6ca08c18799f)

The requested URL begins with `https://giscus.app/zh-cn/widget?origin=`, I tried to remove `zh-cn/` from URL, then the giscus widget can be accessed. Inspired by this discovery, I disabled i18n configs in `./hugo.yaml` and removed i18n folders under `./content` path, the loaded giscus widget proved my guess.

So I think hextra may doesn't consider interactions between i18n folder and other functions of hextra.

You can see these bugs in [my blog](https://librehugohu.github.io/) and find code from [blog repos](https://github.com/librehugohu/librehugohu.github.io).

imfing:
Thanks for reporting, I will look into this

imfing:
I wasn't able to reproduce this issue:

![image](https://github.com/imfing/hextra/assets/5097752/5199e13c-16f3-4c04-b8a3-e429e58eac94)

Could you provide more details?

librehugohu:
I noticed that your Request URL contains `zh-CN`, while mine is `zh-cn`. The Giscus widget will returns 404 with `zh-cn` but works correctly with `zh-CN`.

I don't understand why.

imfing:
@librehugohu did you set the lang code like:

https://github.com/imfing/hextra/blob/7191e259584e339b499a63e2e60cb6d818222e18/exampleSite/hugo.yaml#L27

librehugohu:
Oops, I didn't. Giscus widget appears after setting this option.

PR #358, opened 2024-04-15
Fix: giscus.html fails when country code is specified.

dogweather:
The problem is that the Giscus API fails when given `data-lang` of something like `de-DE`:

Screenshot 2024-04-15 at 15 40 00

This fix uses just the language tag, not the whole language code: https://gohugo.io/methods/site/language/#methods

With this fix:

Screenshot 2024-04-15 at 15 47 13

### DIFF
--- a/layouts/partials/components/giscus.html
+++ b/layouts/partials/components/giscus.html
@@ -1,4 +1,4 @@
-{{- $lang := site.Language.LanguageCode | default `en` -}}
+{{- $lang := site.Language.Lang | default `en` -}}
 
 {{- with site.Params.comments.giscus -}}
 <script>


imfing:
Thanks for the PR. It makes sense. 

I checked the [list of available locales](https://github.com/giscus/giscus/tree/main/locales) on giscus and noticed that `zh-CN` and `zh-TW` use mixed casing. 
With this change, the giscus will now use the lowercase zh-cn in our example site, which can be seen here:
https://github.com/imfing/hextra/blob/be7e0d3f40f8f39d6323f13c3ae0ef253499ed5d/exampleSite/hugo.yaml#L25

Could you try changing `zh-cn` to `zh-CN` in the above link to see if everything still works correctly?

dogweather:
> With this change, the giscus will now use the lowercase zh-cn in our example site, which can be seen here:
> 
> https://github.com/imfing/hextra/blob/be7e0d3f40f8f39d6323f13c3ae0ef253499ed5d/exampleSite/hugo.yaml#L25
> 
> Could you try changing `zh-cn` to `zh-CN` in the above link to see if everything still works correctly?

I'm confused, sorry. With my change, it will just use the first two letters of the code. Here's my debug output I verified this with:

```javascript
".Language.Lang": "zh"
".Language.LanguageCode": "zh-CN"
```

And that currently breaks because giscus doesn't have i18n support for simple `zh`.

dogweather:
I added special case handling for Chinese. I tested it with:

```yaml
languageCode: zh
```

```yaml
languageCode: zh-cn
```

```yaml
languageCode: zh-CN
```

imfing:
Awesome,appreciate it 👍
I was asking because of this issue https://github.com/imfing/hextra/issues/231

Will review it later today

layouts/partials/components/giscus.html
```diff
@@ -1,4 +1,26 @@
...
+
+{{ $sl := site.Language }}
+{{ $giscus_lang := $sl.Lang | default `en` }}
+
+{{/* 
+  Special case for Chinese.
+  Giscus uses the geophraphical language code for these.
+  See: https://github.com/giscus/giscus/tree/main/locales
+  */}}
+{{ if eq $giscus_lang "zh" }}
```
imfing, line 11: In this case, `$giscus_lang` could also be `zh-cn`, so it's better to use `hasPrefix` instead of `eq`.

PR #359, opened 2024-04-15
feat: Support Hero Icons in the nav bar menu

dogweather:
A Hugo `menus` config with `params.hero_icon`:

```yaml
menus:
  main:
  - identifier: discussions
    name: Forum
    url: https://github.com/dogweather/forkful/discussions
    weight: 1
    params:
      hero_icon: chat-alt-2
```

...displays the icon next to the text.

Screenshot 2024-04-15 at 16 24 41

---

I used GitHub and OpenAI as design examples for size and distance to the text. Here are more examples of combinations of icons/non-icons:

Screenshot 2024-04-15 at 16 24 16

Screenshot 2024-04-15 at 16 23 52

This code is running live here: https://forkful.ai/en/

### DIFF
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -56,6 +56,12 @@
             {{ if $external }}target="_blank" rel="noreferer"{{ end }}
             class="hx-text-sm contrast-more:hx-text-gray-700 contrast-more:dark:hx-text-gray-100 hx-relative -hx-ml-2 hx-hidden hx-whitespace-nowrap hx-p-2 md:hx-inline-block {{ $activeClass }}"
           >
+            {{ with .Params.hero_icon -}}
+              <span style="float: left; padding-right: 0.3em; padding-top: 2px; opacity: 0.75">
+                {{ partial "utils/icon.html" (dict "name" . "attributes" "height=18") }}
+              </span>
+            {{- end }}
+
             <span class="hx-text-center">{{ or (T .Identifier) .Name | safeHTML }}</span>
           </a>
         {{- end -}}


layouts/partials/navbar.html
```diff
@@ -56,6 +56,12 @@
             {{ if $external }}target="_blank" rel="noreferer"{{ end }}
             class="hx-text-sm contrast-more:hx-text-gray-700 contrast-more:dark:hx-text-gray-100 hx-relative -hx-ml-2 hx-hidden hx-whitespace-nowrap hx-p-2 md:hx-inline-block {{ $activeClass }}"
           >
+            {{ with .Params.hero_icon -}}
```
imfing, line 59:
```suggestion
            {{ with .Params.icon -}}
```

I'd prefer `icon` over `hero_icon`

layouts/partials/navbar.html
```diff
@@ -56,6 +56,12 @@
...
+              <span style="float: left; padding-right: 0.3em; padding-top: 2px; opacity: 0.75">
```
imfing, line 60: Can we use [Tailwind CSS](https://tailwindcss.com/docs/utility-first) classes here?

imfing (changes requested):
Thank you, this is a nice feature.
Let's enhance the implementation slightly.

layouts/partials/navbar.html
dogweather: Ah, but `.icon` is what sets it as icon _only_, I believe. That's why I went with `hero_icon`. E.g.:

```yaml
  - name: GitHub
    weight: 5
    url: "https://github.com/dogweather/forkful"
    params:
      icon: github
```

layouts/partials/navbar.html
dogweather: I bet we could. I was hoping you'd know what they are. :-)

dogweather: I can begin converting them.

layouts/partials/navbar.html
dogweather: I don't know how much you want to refactor this. Personally, I do "making illegal states impossible". We'd use `type` for these as well (it's already used for `type: search`. That could look like this:

```yaml
- name: github
  params:
    type: icon
    icon: github
- name: forum
  params:
    type: url
    icon: chat-alt-2
```

imfing: I like the idea of using an extra `type` there, let's consider these cases

```yaml
# Icon only, for backward compatibility
- name: github
  url: https://github.com/
  params:
    icon: github

# Link with icon, must set `type: link` and `icon`
- name: forum
  params:
    type: link
    icon: chat-alt-2

# Search bar only
- name: Search
  params:
    type: search
    icon: chat-alt-2 # undefined behavior, currently ignored
```

In the future, we may extend the `type` to other widgets, for example, light/dark/system theme toggle switch:
```yaml
- name: Theme toggle
  params:
    type: theme-toggle
```

PR #367, opened 2024-04-21
feat: add support for `breadcrumbs` param in front matter for showing/hiding breadcrumbs on specific pages

ggwicz:
Hi there! Hextra is awesome and I wanted to propose a feature that I've found useful recently: **the ability to hide breadcrumbs on specific pages.**

This has been especially useful on "index" pages, for example, where even with a custom `linkTitle` there's just not much value added from the breadcrumb.

I've also included some documentation updates for breadcrumbs in general—and for this `hideBreadcrumbs` option specifically—as I wasn't finding that information elsewhere.

### Before

|`/docs`|`/docs/processes/story-points`|
|---|---|
| At my Docs index page (`/docs`), the title `Documentation` being repeated twice by the auto-generated breadcrumbs isn't really helpful. hextra-breadcrumbs-before-A | But it is, of course, still helpful once we get into the actual sub-pages, like `/docs/processes/story-points` in this example: hextra-breadcrumbs-before-B |

### After

With this new `hideBreadcrumbs` param, you can have the best of both worlds.

I add it to the front matter of `/content/docs/_index.md`:
```yaml
---
hideBreadcrumbs: true
title: Documentation
weight: 10
cascade:
  type: docs
  sidebar:
    open: true
---
```

And then no longer have it on the `/docs` page, but **all the other breadcrumbs generation continues to work exactly the same.**

|`/docs`|`/docs/processes/story-points`|
|---|---|
| hextra-breadcrumbs-after-A | hextra-breadcrumbs-after-B |

### DIFF
--- a/exampleSite/content/docs/guide/organize-files.md
+++ b/exampleSite/content/docs/guide/organize-files.md
@@ -84,7 +84,7 @@ weight: 2
 
 ## Breadcrumb Navigation
 
-Breadcrumbs are auto-generated based on the directory structure of `/content`. 
+Breadcrumbs are auto-generated based on the directory structure of `/content`.
 
 For example, consider the file structure [demonstrated above](#directory-structure). Given that structure, the breadcrumbs atop the page at `/docs/guide/organize-files/` would appear automatically as follows:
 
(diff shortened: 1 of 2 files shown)


imfing (changes requested):
Hi @ggwicz, thank you for the PR. I was also thinking about adding this recently.

I wonder if changing the flag name from `hideBreadcrumbs` to simply:

```yaml
---
breadcrumbs: true # or false
---
```

would make more sense, similar to the `toc` flag.

ggwicz:
Yeah good call @imfing! Fixed!

PR #371, opened 2024-04-24
docs: fix deploy-site zh-cn lang filename

kowyo:
I noticed that there are two pages with the same title, but they are in different languages. It is a file-naming error, it is fixed in this PR.


imfing:
nice catch, thank you!

PR #374, opened 2024-04-26
docs: add i18n README links

longlin10086:
Add links to different README, so readers can find it easier.


imfing (approved):
@longlin10086 thank you 👍

PR #394, opened 2024-05-21
docs(fix): correct a mistake in markdown guide

lulunac27a:
Fix code mistake by making sure backticks are inserted at start and end of example text


imfing (approved):
thanks

PR #404, opened 2024-05-27
fix: use `hugo.IsMultilingual` to replace deprecated `.Site.IsMultiLingual`

deining:
When running the example site, a deprecation warning is shown:

```
INFO  deprecated: .Site.IsMultiLingual was deprecated in Hugo v0.124.0 and will be removed in a future release. Use hugo.IsMultilingual instead.
```

This PR fixes this warning.


imfing (approved):
@deining Thanks for the fix, and appreciate updating the docs

PR #405, opened 2024-05-29
i18n: Add Hebrew translation

ShaharNaveh:
Hey, added Hebrew translation:)


imfing (approved):
@ShaharNaveh thank you!

PR #418, opened 2024-06-28
DOC: Add pdf shortcode documentation 

ShaharNaveh:
Added missing documentation for the pdf shortcode.

---

I was conflicted whether to use an external URL or reference a path for the example.
On one hand it's more stable to include a sample pdf file in the repo and referencing it from the example, but on the other hand adding a few bytes for each `git clone` from now on can hurt the performance of CI/CD pipelines.

Ended up with using an external URL, but it won't be difficult changing it.

Another more "stable" option is to upload the sample pdf file to a github comment and to reference that URL instead of a completely external one.

### DIFF
--- a/exampleSite/content/docs/guide/shortcodes/pdf.md
+++ b/exampleSite/content/docs/guide/shortcodes/pdf.md
@@ -1,0 +1,18 @@
+---
+title: PDF
+---
+
+## Example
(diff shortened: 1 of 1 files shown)


imfing:
Thanks, will take a look

imfing:
thanks, I included your change in #441 where I put the pdf shortcode in `Others` section.

PR #427, opened 2024-07-24
feat: add tag support for card shortcode

juliamrch:
## Overview

At Clever Cloud, we customized the card shortcode to accept tags, in order to mark the status of some products (beta, tech preview, etc). See [this page](https://developers.clever-cloud.com/doc/addons/) as example. We thought you could be interested in the feature so we took the time to implement it natively in the theme. 

## Tag features

Tags can be added to cards with custom text, on default color (`gray`) or custom ones (`blue`, `yellow`, `red`).  

Capture d’écran 2024-07-24 à 12 46 54

Capture d’écran 2024-07-24 à 12 47 05

## Changes

We followed the same logic applied in the [`callout` shortcode](https://github.com/imfing/hextra/blob/main/layouts/shortcodes/callout.html) to ensure coherence across the codebase. This PR brings the following changes:

- A new parameter in the card shortcode
- A smaller font available in CSS and used for the tag (`hx-text-xxs`)
- A new class named `hx-tag`
- Updated documentation in `/exampleSite` : [see the preview](https://deploy-preview-427--hugo-hextra.netlify.app/docs/guide/shortcodes/cards/)

Lmk if you like the idea or if you hate it (both are ok).


imfing (approved):
Thanks, this looks great :)

I will create a follow-up PR to extract the implementation to a new partial so it can be used in other places. Also, I plan to change the interface slightly - instead of `tagColor` I would prefer to use `tagType` to be consistent with the callout component

PR #435, opened 2024-08-14
fix: remove deprecated google analytics reference in head.html

eskopp:
Fix:

```
ERROR deprecated: .Site.GoogleAnalytics was deprecated in Hugo v0.120.0 and will be removed in Hugo 0.133.0.
 Use .Site.Config.Services.GoogleAnalytics.ID instead.
```

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -30,12 +30,11 @@
   {{ partialCached "head-css.html" . }}
 
 
-  <!-- Google Analytics -->
-  {{- if and (eq hugo.Environment "production") (or .Site.GoogleAnalytics .Site.Config.Services.GoogleAnalytics.ID) }}
-    <link rel="preconnect" href="https://www.googletagmanager.com" crossorigin />
-    {{ partial "google-analytics.html" . }}
-  {{- end }}
-
+<!-- Google Analytics -->
+{{- if and (eq hugo.Environment "production") .Site.Config.Services.GoogleAnalytics.ID }}
+  <link rel="preconnect" href="https://www.googletagmanager.com" crossorigin />
+  {{ partial "google-analytics.html" . }}
+{{- end }}
 
   <script>
     /* Initialize light/dark mode */


layouts/partials/head.html
```diff
@@ -30,12 +30,11 @@
...
-  <!-- Google Analytics -->
-  {{- if and (eq hugo.Environment "production") (or .Site.GoogleAnalytics .Site.Config.Services.GoogleAnalytics.ID) }}
-    <link rel="preconnect" href="https://www.googletagmanager.com" crossorigin />
-    {{ partial "google-analytics.html" . }}
-  {{- end }}
-
+<!-- Google Analytics -->
+{{- if and (eq hugo.Environment "production") .Site.Config.Services.GoogleAnalytics.ID }}
+  <link rel="preconnect" href="https://www.googletagmanager.com" crossorigin />
+  {{ partial "google-analytics.html" . }}
```
imfing, line 36: also need to update this place

imfing (changes requested):

layouts/partials/head.html
eskopp: Yes, the change is completely sufficient. Previously there was only a warning. Now, however, there is an error and the project crashes. The change is therefore necessary.

See here:
 -  https://gohugo.io/methods/site/googleanalytics/

imfing: i mean in this file https://github.com/imfing/hextra/blob/main/layouts/partials/google-analytics.html

eskopp: To be fair, I only made the change I described. I then tested my private project. As there was no warning and no error, the issue was closed for me.

imfing: @eskopp as I pointed out above, this PR didn't address all the locations where `Site.GoogleAnalytics` is referenced, thus it will not close the issue for me. 

anyway, I will open a following PR to address this.

PR #446, opened 2024-08-24
feat: add tooltips to cards

joe128:
#### Added 2 Params to shortcode `card`:

- linktooltip
Hextra: Tooltip for Link
  
- tagtooltip
Hextra: Tooltip for Tag

#### open TODOs
-  Add Documentation to `fa` and `zh-cn`
- Maybe Styling in Dark-Mode

### DIFF
--- a/exampleSite/content/docs/guide/shortcodes/cards.md
+++ b/exampleSite/content/docs/guide/shortcodes/cards.md
@@ -7,8 +7,8 @@ linkTitle: Cards
 
 {{< cards >}}
   {{< card link="../callout" title="Callout" icon="warning" >}}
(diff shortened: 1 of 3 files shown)


imfing (changes requested):
Hi @joe128 thank you for the PR! 

I appreciate the effort you’ve put into this. but I’m not sure if using the title attribute of a link as a tooltip is the best approach. 

While allowing users to set a link title for a card is a good idea, I believe we should avoid using title as a tooltip as they serve different purposes.

joe128:
The problem is that there is no "standard" for tooltips but most browsers interpret the title as a tooltip.
So, what do you think about using a "data-tooltip" ([HTML data-* Attribute](https://www.w3schools.com/tags/att_data-.asp)) for the two elements?

PR #456, opened 2024-09-22
docs: fix incorrect icon in example

icannotfly:
The `tag` icon was incorrectly shown in the example as being `warning`, which can be confusing when trying to correlate the two.


imfing (approved):
@icannotfly thank you!

PR #470, opened 2024-10-10
feat: add customizable primary lightness support

Gr3q:
Lightness is adjusted in a way that `--primary-lightness` becomes the midpoint of the 0-100% range and the calculated result will never go out of that range, it's not a simple addition. 95% lightness is still almost white and 5% is still almost completely black.

This seem to work pretty well with the exception that hextra seem to apply 5% lower lightness in most prominent places (links and buttons), so you need to consistently add +5% lightness to your `--primary-lightness` value but I'm not going to fix that in this PR.

Very high and very low `--primary-lightness` values are not recommended, but they do work.

Something to visualize the algorithm on a 0-100 scale: https://www.desmos.com/calculator/lsdmffszmy (`Y` is base color lightness, `x` is the target lightness)

### DIFF
--- a/assets/css/compiled/main.css
+++ b/assets/css/compiled/main.css
@@ -1006,18 +1006,18 @@ video {
 }
 .hx-bg-primary-100 {
   --tw-bg-opacity: 1;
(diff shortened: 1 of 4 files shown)


imfing:
thank you for the PR, will take a look at it when i get chance

PR #471, opened 2024-10-12
feat: scroll selected sidebar entry into view

### LINKED ISSUE #395: Make selected entry in sidebar always visible
**Feature Description**

Currently, for sidebars with expanded menus or with more entries than are visible in the viewport, selecting an entry below the viewport doesn't scroll the sidebar accordingly. Thus, the users have to scroll again if they want to see the current selected entry.

The sidebar could have the current selected entry either scrolled to the top, or scrolled to the middle.

**Problem/Solution**

My current problem is that https://wiki.hyprland.org uses expanded menus, which make the sidebar larger than the viewport. This is mostly fine, until you get to the lower entries in the sidebar, such as https://wiki.hyprland.org/Hypr-Ecosystem/. If you click on the link, you can see that the sidebar isn't scrolled to have the selected entry within view.

**Alternatives Considered**

I have considered not having the menus open at all, but after an internal discussion we decided to keep the behavior as is, due to users having a harder time reading the docs when needing to click more menus.

**Additional Context**

How the sidebar currently works:
image

How it should work instead:
image

yuri1969:
Long sidebars did not scroll to show the selected entry. This made working with such long sidebars quite confusing.

QAed using 10+ copies of `exampleSite/content/docs/guide/shortcodes/callout.md`.

Fixes #395

### DIFF
--- a/assets/js/sidebar.js
+++ b/assets/js/sidebar.js
@@ -1,4 +1,9 @@
 document.addEventListener("DOMContentLoaded", function () {
+  scrollToActiveItem();
+  enableCollapsibles();
+});
+
+function enableCollapsibles() {
   const buttons = document.querySelectorAll(".hextra-sidebar-collapsible-button");
   buttons.forEach(function (button) {
     button.addEventListener("click", function (e) {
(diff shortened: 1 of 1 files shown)


imfing (approved):
@yuri1969 looks great, thanks for the contribution!

Issue #469, opened 2024-10-08
Display page breadcrumbs in search results, not just the page title


Gr3q:
**Feature Description**

Currently in search results are grouped by page. Only the page title is displayed for each group at the moment. It could show the full breadcrumb instead. e.g. if the page is called "Devices", it should be called "Docs -> Devices" .

**Problem/Solution**

Without the breadcrumbs it makes search results ambiguous to know which section that page belongs to. This is an even bigger problem if you have multiple pages with the same title.

**Alternatives Considered**

I can change the title to be longer and include parts of the hierarchy, but in that case the sidebar title gets long too.

PR #473, opened 2024-10-14
feat: show breadcrumbs in search results

Gr3q:
* Resolve https://github.com/imfing/hextra/issues/469

Shows breadcrumbs instead of the page title for each section of the search results.

For some reason `data` can be incomplete on [our site](https://github.com/netduma/knowledge-base), I need to figure out why exactly.

### DIFF
--- a/assets/js/flexsearch.js
+++ b/assets/js/flexsearch.js
@@ -200,7 +200,7 @@ document.addEventListener("DOMContentLoaded", function () {
       cache: 100,
       document: {
         id: 'id',
-        store: ['title'],
+        store: ['title', 'crumb'],
         index: "content"
       }
     });
@@ -210,7 +210,7 @@ document.addEventListener("DOMContentLoaded", function () {
       cache: 100,
       document: {
         id: 'id',
-        store: ['title', 'content', 'url', 'display'],
+        store: ['title', 'content', 'url', 'display', 'crumb'],
         index: "content",
         tag: 'pageId'
       }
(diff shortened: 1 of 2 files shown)


imfing:
> For some reason data can be incomplete on [our site](https://github.com/netduma/knowledge-base), I need to figure out why exactly.

might be caused by the parsing logic, i suggest you take a look at: https://github.com/imfing/hextra/blob/main/layouts/partials/utils/fragments.html

Gr3q:
I had a look, it's this line: https://github.com/imfing/hextra/blob/97ea67198b2f5439b5ba727e831ea4f7d30a0df2/assets/json/search-data.json#L10

It excludes pages from search with no content. There are 2 options:
* Empty pages should not be excluded from search (can still be explicitly excluded with `excludeSearch`)
* Should just warn in console (like now in the PR) for both excluded/empty content pages that have children.

To be honest pages that have children should not be excluded with `excludeSearch` nor they should be empty most of the time.

It's up to you which option you choose for this PR - for our page I'll have to add content for the relevant pages (as they should have content)

imfing:
@Gr3q thanks the PR overall makes sense to me. 

re: page exclusion that you mentioned

> It excludes pages from search with no content. 

I kinda forget what was the reason to put that line to exclude page without content. Maybe we could remove that line since we have `excludeSearch`?

Gr3q:
> I kinda forget what was the reason to put that line to exclude page without content. Maybe we could remove that line since we have excludeSearch?

Can do, I checked before and it doesn't seem to break search in any way, and you can search for the title of empty pages.

imfing (approved):
looks good, thanks!

PR #474, opened 2024-10-14
feat: make cards as partial

### LINKED ISSUE #464: Make card(s) usable as partial
**Feature Description**

I would like to make either a list template or a shortcode (or both) that displays a collection of pages as cards.

**Problem/Solution**

If I'm understanding things correctly, you can't call `layouts/shortcodes/cards.html` from within a partial, so it's functionality would need to be moved to a partial and the shortcode should become a very thin wrapper around the partial. I think, in general, this is a good pattern to follow. The Relearn theme seems to do it a lot.

**Alternatives Considered**

If found #439 but I don't think the workaround posted there works anymore.

imfing:
closes #464

### DIFF
--- a/layouts/partials/shortcodes/cards.html
+++ b/layouts/partials/shortcodes/cards.html
@@ -1,0 +1,6 @@
+{{- $cols := .cols | default 3 -}}
+{{- $content := .content -}}
+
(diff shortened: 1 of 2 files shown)


imfing:
https://deploy-preview-474--hugo-hextra.netlify.app/docs/guide/shortcodes/cards/ - everything seems fine

Issue #490, opened 2024-11-03
Make selected entry in sidebar always visible, same problem like issues#395


xue68:
**Feature Description**

Make selected entry in sidebar always visible. same problem like [#395](https://github.com/imfing/hextra/issues/395). I want to try a new solution below. The old solution pr[#471](https://github.com/imfing/hextra/pull/471) always put the active item in the top of the sidebar list. When I merged the pr, I found it looks like strange. After I saw the solution of vitpress, I think saving the bar position is better.

**Problem/Solution**

This is the new sidebar.js

```javascript
document.addEventListener("DOMContentLoaded", function () {
  restoreSidebarPosition();
  enableCollapsibles();
});

function enableCollapsibles() {
  const buttons = document.querySelectorAll(".hextra-sidebar-collapsible-button");
  buttons.forEach(function (button) {
    button.addEventListener("click", function (e) {
      e.preventDefault();
      const list = button.parentElement.parentElement;
      if (list) {
        list.classList.toggle("open")
      }
    });
  });
}

function saveSidebarPosition() {
  const sidebarScrollbar = document.querySelector("aside.sidebar-container > .hextra-scrollbar");
  if (sidebarScrollbar) {
    const scrollPosition = sidebarScrollbar.scrollTop;
    sessionStorage.setItem('sidebarScrollPosition', scrollPosition);
  }
}

function restoreSidebarPosition() {
  const sidebarScrollbar = document.querySelector("aside.sidebar-container > .hextra-scrollbar");
  if (sidebarScrollbar) {
    const savedPosition = sessionStorage.getItem('sidebarScrollPosition');
    
    if (savedPosition !== null) {
      sidebarScrollbar.scrollTo({
        top: parseInt(savedPosition),
        behavior: 'instant'
      });
    }

    sidebarScrollbar.addEventListener('scroll', function() {
      saveSidebarPosition();
    });
  }
}
```

**Alternatives Considered**

null

**Additional Context**

There may be a lot of problems with this place, please advise.

imfing:
Hi @xue68 thank you for the feedback. Would you mind opening a pull request so we can preview the above changes you proposed

xue68:
ok, the new pr #491 has been commit.

PR #492, opened 2024-11-03
fix: restore the sidebar position to make the selected entry showed

xue68:
save and restore the sidebar position to make the selected entry showed. #490

### DIFF
--- a/assets/js/sidebar.js
+++ b/assets/js/sidebar.js
@@ -1,7 +1,30 @@
 document.addEventListener("DOMContentLoaded", function () {
-  restoreSidebarPosition();
-  enableCollapsibles();
+  initializeSidebar();
 });
+
+function initializeSidebar() {
+  const sidebarScrollbar = document.querySelector("aside.sidebar-container > .hextra-scrollbar");
+  if (!sidebarScrollbar) return;
+
+  enableCollapsibles();
+  restoreSidebarPosition(sidebarScrollbar);
+
+  const debouncedSave = debounce((position) => {
+    saveSidebarPosition(position);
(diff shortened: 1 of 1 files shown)


assets/js/sidebar.js
```diff
@@ -16,21 +16,28 @@ function enableCollapsibles() {
...
+    
+    if (savedPosition !== null) {
+      sidebarScrollbar.scrollTo({
+        top: parseInt(savedPosition),
+        behavior: 'instant'
+      });
+    }
+
+    sidebarScrollbar.addEventListener('scroll', function() {
+      saveSidebarPosition();
```
imfing, line 40: I think you need debouncing the event handler.

imfing (changes requested):
btw, can you share the vitepress implementation you mentioned for saving the sidebar scroll position?

xue68:
To be honest, I don't know how Vitepress is implemented, I'm just referring to the effects.

assets/js/sidebar.js
xue68: ok, I have add the new function.

imfing:
There’s an issue after this change: switching between sidebar items would cause a noticeable jittering effect every time.

https://github.com/user-attachments/assets/d8dcb6ff-57a4-404e-873b-bb04741f15fa


Previous:

https://github.com/user-attachments/assets/92c2263d-f2a3-4932-adbc-cf00de8ea480

xue68:
I've noticed this issue, but I've been struggling to figure out how to fix it. After testing, it is now possible to significantly reduce the frequency of occurrences.

Issue #343, opened 2024-03-31
Add light/dark theme switch to navigation


imfing:
### Discussed in https://github.com/imfing/hextra/discussions/341

<div type='discussions-op-text'>

<sup>Originally posted by **lfenzo** March 31, 2024</sup>
Is there any way to move the light/dark theme switch to the menu at the top of the page?</div>

PR #499, opened 2024-11-16
feat: support theme toggle in navbar.

jason-dour:
- Add additional logic for navbar partial.
- Uses a menu item params.type of `theme-toggle`.
- Uses additional params.hideLabel boolean to set hideLabel for the theme-toggle.html partial; defaults to true.

Addresses #343 .

### DIFF
--- a/layouts/partials/theme-toggle.html
+++ b/layouts/partials/theme-toggle.html
@@ -1,3 +1,6 @@
+{{- $navbar := .navbar | default false -}}
+{{- $height := 12 -}}
+{{- if $navbar -}}{{- $height = 24 -}}{{- end -}}
 {{- $hideLabel := .hideLabel | default false -}}
 
 {{- $changeTheme := (T "changeTheme") | default "Change theme" -}}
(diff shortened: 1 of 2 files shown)


jason-dour:
This isn't a perfect PR, as I didn't include any updates to documentation for this feature and instead focused on implementation.

I figured if this looks good, I could add documentation changes to this PR. :)

jason-dour:
Gave theme-toggle partial logic based on navbar boolean, to set size of icon based on navbar or elsewhere.

This does not address the size of the toggle's label. I see the `hx-text-xs` present in theme-toggle. Could possibly add more logic around that.

jason-dour:
Put logic in theme-toggle when in navbar to match size of navbar text.

layouts/partials/theme-toggle.html
```diff
@@ -1,3 +1,10 @@
+{{- $navbar := .navbar | default false -}}
+{{- $height_px := 12 -}}
+{{- $height_tw := "hx-text-xs" -}}
+{{- if $navbar -}}
+  {{- $height_px = 24 -}}
+  {{- $height_tw = "hx-text-sm" -}}
+{{- end -}}
```
imfing, line 7: just curious where are `12` and `24` coming from?

imfing (commented):
will take a second look by manually testing it.

layouts/partials/theme-toggle.html
jason-dour: Great question!  I've been offline for a few weeks, and honestly, I'd have to say it is likely a best estimate based on testing that made things look "right."

If those values can be computed from other inputs in the theme configuration, I'm certainly willing to explore that.  Do you have any guidance on how we might calculate a sane value for this?

jason-dour:
> @jason-dour @imfing What is the status here, why is this not being merged? Is there anything I can do to help?

Honestly, I submitted this in hopes of helping, and made some assumptions (see above comments about icon sizes) based on limited knowledge of the internals of hextra.

I chose 12 and 24 because it seemed "right," and was asked about it, and quite frankly I've been too busy to come back to it.

I think it is functional, and I know of no other way to set the size of the icons in a more theme compliant way. Open to ideas.

PR #509, opened 2024-12-07
fix: rerender mermaid diagrams on theme change

torbjornbp:
Make mermaid render diagrams on manual toggle between themes, after initial page load.

The original code renders the correct theme only on the initial page load. Toggling to a different theme after initial page load does not change the mermaid theme, which leads to light grey text on white background etc.


imfing (approved):
Thank you. Looks great!

PR #518, opened 2024-12-16
feat: support custom footer section

Gr3q:
See the documentation changes for explanation.

Still, I'm not sure if this is a feature that is wanted as is because it forces you to write html.

Unfortunately I couldn't find a way to use a markdown file as a partial instead unless it's like this https://stackoverflow.com/a/39800156.


imfing:
@Gr3q thank you, supporting custom footer is a nice feature to add.

I was thinking we can simplify your approach by adding `custom/footer-end.html`, and then update the `footer.html` to include it:

```diff
     <div class="hx-flex hx-w-full hx-flex-col hx-items-center sm:hx-items-start">
       {{- if (.Site.Params.footer.displayPoweredBy | default true) }}<div class="hx-font-semibold">{{ template "theme-credit" $poweredBy }}</div>{{ end }}
       {{- if .Site.Params.footer.displayCopyright }}<div class="hx-mt-6 hx-text-xs">{{ $copyright | markdownify }}</div>{{ end }}
+      {{ partial "custom/footer-end.html" . }}
     </div>
```

thus, we can use hugo template syntax like this to render markdown from either in the template or load a param from hugo config file

```
{{ "**hello**" | markdownify }}

<!-- or load from your own custom parameter -->
{{ site.Params.footer.customFooter | markdownify }}
```

example hugo.yaml

```yaml
  footer:
    enable: true
    displayCopyright: false
    displayPoweredBy: false
    width: normal
    customFooter: "**custom footer**" # you can change or add your own param
```

output:

image

Gr3q:
> thus, we can use hugo template syntax like this to render markdown from either in the template or load a param from hugo config file

`markdownify` only supports markdown features as long as it resolves to a single paragraph, so I can't do tables headings etc if I want to (I don't, but some other layout would come handy). And I can do the same with the copyright string.

> I was thinking we can simplify your approach by adding custom/footer-end.html, and then update the footer.html to include it:

I started with that and I found the padding around too limiting. Still it's a good idea, except the footer should be sandwiched between the copyright/powered by text and the theme buttons (if it's a page that has the theme buttons there) separated by dividers.

The reason is that most websites have the copyright section on the very bottom but I can't link to some dataset to back up that claim.

Gr3q:
So I can change the PR to complement the copyright section instead of replacing it.

In that case I'll try to upstream some other things in more PRs like changing the footer background/text color and the copyright section layout (centered and in a row instead of having a column)

layouts/partials/footer.html
```diff
@@ -24,14 +24,23 @@
...
-    <div class="hx-flex hx-w-full hx-flex-col hx-items-center sm:hx-items-start">
-      {{- if (.Site.Params.footer.displayPoweredBy | default true) }}<div class="hx-font-semibold">{{ template "theme-credit" $poweredBy }}</div>{{ end }}
-      {{- if .Site.Params.footer.displayCopyright }}<div class="hx-mt-6 hx-text-xs">{{ $copyright | markdownify }}</div>{{ end }}
+
+  {{- if .Site.Params.footer.customSection -}}
+    <div
+      class="{{ $footerWidth }} hx-pl-[max(env(safe-area-inset-left),0rem)] hx-pr-[max(env(safe-area-inset-right),0rem)] hx-text-gray-600 dark:hx-text-gray-400"
+    >
+      {{ partial "custom/footer.html" (dict "context" . "footerButtonColor" "hx-rounded-md hx-transition-colors hover:hx-bg-gray-100 hover:hx-text-gray-900 dark:hover:hx-bg-primary-100/5 dark:hover:hx-text-gray-50" ) }}
+    </div>
```
imfing, line 33: `footerButtonColor` is not necessary here right?

exampleSite/content/docs/advanced/customization.md
```diff
@@ -62,6 +62,27 @@ You may add custom scripts to the end of the head for every page by adding the f
...
 ```
 
+## Custom Footer Section
+
+You can add a custom section the footer between the language/theme buttons and the Copyright/Powered By section by creating a file `layouts/partials/custom/footer.html` in your site __and__ setting `customSection: true` in the `footer` section of your `hugo.yaml`:
+
+```yaml {filename="hugo.yaml"}
+ footer:
+    enable: true
+    customSection: true
```
imfing, line 72: I would avoid adding a new flag if possible

Gr3q: I can remove it but then I don't know when I should show the separator and when I shouldn't.

layouts/partials/footer.html
Gr3q: Not really, but the other option is to dig through the hextra codebase to match it. I'll remove it.

exampleSite/content/docs/advanced/customization.md
Gr3q: Actually, I just remove the separator, people can add it back in their custom footer.

Gr3q:
I also added a breaking change to only show footer section for copyright and PoweredBy if they are enabled

Out of the box the footer is there unless people explicitly disable poweredby and the copyright section.

layouts/partials/footer.html
```diff
@@ -24,14 +26,21 @@
...
   {{- end -}}
-  <div
-    class="{{ $footerWidth }} hx-mx-auto hx-flex hx-justify-center hx-py-12 hx-pl-[max(env(safe-area-inset-left),1.5rem)] hx-pr-[max(env(safe-area-inset-right),1.5rem)] hx-text-gray-600 dark:hx-text-gray-400 md:hx-justify-start"
-  >
-    <div class="hx-flex hx-w-full hx-flex-col hx-items-center sm:hx-items-start">
-      {{- if (.Site.Params.footer.displayPoweredBy | default true) }}<div class="hx-font-semibold">{{ template "theme-credit" $poweredBy }}</div>{{ end }}
-      {{- if .Site.Params.footer.displayCopyright }}<div class="hx-mt-6 hx-text-xs">{{ $copyright | markdownify }}</div>{{ end }}
+    <div
+      class="{{ $footerWidth }} hx-pl-[max(env(safe-area-inset-left),0rem)] hx-pr-[max(env(safe-area-inset-right),0rem)] hx-text-gray-600 dark:hx-text-gray-400"
+    >
```
imfing, line 31: I'd prefer keep the left and right padding, e.g.

```
<div class="{{ $footerWidth }} hx-mx-auto hx-pl-[max(env(safe-area-inset-left),1.5rem)] hx-pr-[max(env(safe-area-inset-right),1.5rem)] hx-text-gray-600 dark:hx-text-gray-400" >
```

to make this section styling more consistent

layouts/partials/footer.html
```diff
@@ -24,14 +26,21 @@
...
+      {{ partial "custom/footer.html" (dict "context" . "aboveVisible" $footerSwitchesVisible "belowVisible" $copyrightSectionVisible ) }}
```
imfing, line 32: what are these variables used for exactly?

imfing (changes requested):
Let's keep the left and right padding. 

I tried your change with a custom html footer in `custom/footer.html`

```
<div style="display: flex; justify-content: space-between; padding-top: 20px;">
  <div style="flex: 1;">
      <h4 style="margin-bottom: 10px; font-weight: 600;">About Us</h4>
      <p>Learn more about our company.</p>
  </div>
  <div style="flex: 1;">
      <h4 style="margin-bottom: 10px; font-weight: 600;">Services</h4>
      <ul style="list-style: none; padding: 0;">
          <li style="margin: 5px 0;">Web Design</li>
          <li style="margin: 5px 0;">SEO</li>
          <li style="margin: 5px 0;">Consulting</li>
      </ul>
  </div>
  <div style="flex: 1;">
      <h4 style="margin-bottom: 10px; font-weight: 600;">Contact</h4>
      <p>Email: contact@example.com</p>
      <p>Phone: 123-456-7890</p>
  </div>
</div>
```
image

It works pretty well

layouts/partials/footer.html
Gr3q: Fair, but I will add a custom class so people can customize its style

imfing (approved):
thanks, awesome work

PR #523, opened 2024-12-22
feat: support link in code block title

strowk:
This introduces new attribute `filename_uri_base` that adds to codeblock rendering a feature allowing to render a link from the code block title to wherever the sources could be looked up.

image

UPD: you can see it here: https://deploy-preview-523--hugo-hextra.netlify.app/docs/guide/syntax-highlighting/#link-to-file

### DIFF
--- a/exampleSite/content/docs/guide/syntax-highlighting.md
+++ b/exampleSite/content/docs/guide/syntax-highlighting.md
@@ -40,6 +40,23 @@ def say_hello():
     print("Hello!")
 ```
 
+### Link to file
+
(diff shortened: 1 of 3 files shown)


imfing:
@strowk thanks a lot! I like the idea. I’ll take a closer look at it soon.

imfing:
@strowk I simplified the implementation a bit by using more hugo builtin methods

also renamed the param `filename_uri_base` to `base_url` (let me know what your thoughts are) cuz it's a bit lengthy imo

Issue #380, opened 2024-05-05
Search bar in left sidebar menu


trwnh:
**Feature Description**

Support having a search bar in the left sidebar menu

**Problem/Solution**

The search bar is only supported currently in the top navbar. It would be nice to have the search bar in the left sidebar menu instead.

**Alternatives Considered**

None currently, except for using a top navbar search bar

imfing:
Thanks, I think this is a nice feature to add :)

PR #534, opened 2025-01-03
feat: Add search as a shortcode

Gr3q:
Again, I'm not sure if it's particularly useful, but I opened it because maybe it's desired. I wish I could make the shortcode more convenient. 

This is still a site-wide searchbar, but maybe it can be scoped easily if the base url is passed in as a param down into the search js "class", then filter the pages in `preloadIndex`

Maybe this is not even the best way to do this, I'm not sure.

Changes:
* flexsearch and search element must be unique, so added `name` as a required parameter
* Made keybinding optional so it can be controlled in case you want multiple search elements
* added `not-prose` to some search elements to prevent content styling them over incorrectly
* Due to the first change each element can be styled individually

Known issues:
* `name` is required, but it could be generated (not random!)
* Search result styling was made for the nevbar, so it's wrong everywhere else (`hx-max-w-[min(calc(100vw-2rem),calc(100%+20rem))]`), the `+20rem` is the problem, so currently manual adjustment is needed in `custom.css`.


imfing:
thanks! I think this is very nice to have, which helps address #380 more easily

Gr3q:
I take recommendations on how to deal with the `search-results` width and position correctly.

My own solution was to set it's width to `100%`, it doesn't extend over on either side, but for smaller search bars it will be to small. Plus if you want it to extend and it's on the left side the absolute positioning needs to be adjusted as well.

I'm not sure I can handle extending and positioning without Javasript.

PR #538, opened 2025-01-05
refactor: use generic .Process for card processing and allow bundled images

### LINKED ISSUE #103: Use generic `Process` from Hugo v0.119.0 for card image processing
[Hugo v0.119.0](https://github.com/gohugoio/hugo/releases/tag/v0.119.0) introduced a new function `Process` that is more general

so we don't need the if block in the `card.html` image processing logic:
https://github.com/imfing/hextra/blob/main/layouts/shortcodes/card.html#L15

jzerfowski:
This PR changes the processing of the given card $image to the more generic [.Process](https://gohugo.io/content-management/image-processing/#image-processing-options) method available from hugo v0.119.0.
It allows to process both global images from assets as well as local/page bundle images.

resolves #103

### DIFF
--- a/layouts/shortcodes/card.html
+++ b/layouts/shortcodes/card.html
@@ -38,4 +38,4 @@
   "tag"         $tag
   "tagType"     $tagType
   )
--}
+-}}
--- a/layouts/shortcodes/card.html
+++ b/layouts/shortcodes/card.html
@@ -9,14 +9,17 @@
 {{- $imageStyle := .Get "imageStyle" -}}
 {{- $tag := .Get "tag" -}}
 {{- $tagType := .Get "tagType" -}}
+{{/* hugo process argument string /*}}
 {{- $process := .Get "process" | default "resize 800x webp q80" -}}
 
 {{- with or (.Page.Resources.Get $image) (resources.Get $image) -}}
+  {{/* Retrieve the $image resource from local or global resources */}}
   {{- $processed := .Process $process -}}
   {{- $width = $processed.Width -}}
   {{- $height = $processed.Height -}}
   {{- $image = $processed.RelPermalink -}}
 {{ else }}
+  {{/* Otherwise, use relative link of the image */}}
   {{- if hasPrefix $image "/" -}}
     {{- $image = relURL (strings.TrimPrefix "/" $image) -}}
   {{- end -}}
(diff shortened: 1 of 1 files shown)


layouts/shortcodes/card.html
```diff
@@ -9,34 +9,19 @@
 {{- $imageStyle := .Get "imageStyle" -}}
 {{- $tag := .Get "tag" -}}
 {{- $tagType := .Get "tagType" -}}
+{{/* hugo process argument string /*}}
+{{- $process := .Get "process" | default "resize 800x webp q80" -}}
 
-{{/* Image processing options */}}
-{{- $method := .Get "method" | default "Resize" | humanize -}}
-{{- $options := .Get "options" | default "800x webp q80" -}}
```
imfing, line 15: deleting these will break existing functionality.
to maintain backward compatibility, I recommend setting the `process` parameter default to be `$method $options`

layouts/shortcodes/card.html
```diff
@@ -53,4 +38,4 @@
   "tag"         $tag
   "tagType"     $tagType
   )
--}}
+-}
```
imfing, line 41:
```suggestion
-}}
```

imfing (changes requested):

layouts/shortcodes/card.html
jzerfowski: Thanks, that makes a lot of sense. Fixed it in the new commit. I'm very new to the templating language and syntax, so I'm happy about feedback. There's probably a better way to do the concatenation

layouts/shortcodes/card.html
```diff
@@ -13,30 +13,18 @@
 {{/* Image processing options */}}
 {{- $method := .Get "method" | default "Resize" | humanize -}}
 {{- $options := .Get "options" | default "800x webp q80" -}}
+{{- $process := .Get "process" | default (delimit (slice $method $options) " " " ") -}}
```
imfing, line 16:
```suggestion
{{- $process := .Get "process" | default (printf "%s %s" $method $options) -}}
```

Guess we can just use [printf](https://gohugo.io/functions/fmt/printf/) here?

imfing (commented):
one suggestion, otherwise all good!

Issue #549, opened 2025-01-15
Cards no longer support SVG images


wdullaer:
**Description**

In https://github.com/imfing/hextra/commit/a27f6eef5e2fbdc9570cb05a55befc80427dd346 the code to render images in cards was refactored to use the `.Process` function.
This function only works on raster images, and not on SVG images.

We happened to use SVG images, and trying to use the latest template will result in the following error:

```
Error: error building site: "/mysite/_index.md:6:1": failed to render shortcode "cards": "/mysite/_index.md:1:1": failed to render shortcode "card": failed to process shortcode: "/Users/me/Library/Caches/hugo_cache/modules/filecache/modules/pkg/mod/github.com/imfing/hextra@v0.9.3/layouts/shortcodes/card.html:20:20": execute of template failed: template: shortcodes/card.html:20:20: executing "shortcodes/card.html" at <.Process>: error calling Process: this method is only available for raster images. To determine if an image is SVG, you can do {{ if eq .MediaType.SubType "svg" }}{{ end }}
```

**Steps To Reproduce**

1. Create a card that uses an svg as an image
2. Try to build the site

**Expected Behavior**

The card renders the actual svg image

**Actual Behavior**

Build fails because SVG images are not supported by the `.Process` function

**Environment**

- Hugo Version: v1.140.2
- Browser/OS: MacOS (but irrelevant)
- Theme Version: v0.9.3

**Additional Context**

GelbEinhalb:
I'm not sure if that is a general issue or just a problem on my side, but I think external image links in general don't work anymore. I've switched to a version from before the change since then.
I didn't have time for testing/writing an issue though.

jzerfowski:
I can confirm that apparently my "fix" broke both .svg and remote resources support. Soryr for not testing properly. I wrote some fixes - before I create a pull request, could you check if the changes to `shortcodes/card.html` fix the issues?
@wdullaer @GelbEinhalb 

https://github.com/jzerfowski/hextra/blob/549-fix-card-svg-remote/layouts/shortcodes/card.html

wdullaer:
With that `card.html` the site renders correctly and the svg related issues are gone.

Thanks for the quick follow up!

imfing:
fixed in https://github.com/imfing/hextra/releases/tag/v0.9.4

PR #551, opened 2025-01-17
fix: skip image process on svgs and remote images

jzerfowski:
This commit attempts to fix a bug (#549) introduced by switching to Hugo's generic .Process method (see #103).
It skips processing on remote resources and does not process local .svgs either.


imfing:
@jzerfowski thank you for the fix!

PR #554, opened 2025-01-19
Fix missing doctype error on 404 page with htmltest scan

gaetan-steininger:
When we scan a project using this theme, an error appears: 'missing doctype' on the 404 page. This change fixes the issue.


imfing (approved):
thanks

PR #562, opened 2025-01-23
docs: describe page last modification date

### LINKED ISSUE #548: Documentation is missing for `displayUpdatedDate`
There is no documentation for the `displayUpdatedDate` setting. I discovered it was an option when looking at some of the hextra template files, and noticing there was a last updated date showing on the official doc pages.

Unfortunately it's not documented anywhere, and it took quite a bit of digging to realize it's not enough to set it to true, you also have to set `enableGitInfo` to true.

The Configuration page says to go look at https://github.com/imfing/hextra/blob/main/exampleSite/hugo.yaml, but even that file doesn't have any comments around the setting indicating what it does or the dependency on `enableGitInfo`.

yuri1969:
Added a section to the configuration guide (English only) and annotated the `hugo.yaml` example file.

Fixes #548


imfing (approved):
thank you! @yuri1969

PR #580, opened 2025-02-22
Update title.html

Vraint:
Is it more appropriate to prioritize file names?


Vraint:
A title that is too long can cause the sidebar directory to display multiple lines of text, affecting its appearance

PR #589, opened 2025-02-26
chore: server side rendering of math equations

deining:
This PR brings server side rendering of math equations via hugo's built-in KaTeX rendering engine.

**Advantages:**
- server side rendering (instead of client side rendering)
- no JS needed for math content
- typesetting of formulae using LaTeX standard syntax, no escaping of delimiters needed
- no need to activate math mode in frontmatter any more
- internal rendering engine supports rendering of chemical equations out of the box (as of hugo version 0.144.0)
- no need to include KaTeX dependencies any more

I hope you like this PR.


imfing:
Hi @deining,

Thank you, this is great! I plan to merge it in the next minor version.

I may follow up separately with a small PR to make the katex CDN configurable

Issue #292, opened 2024-02-15
Allow changing background color


Visual-Dawg:
**Feature Description**

Currently it is not possible to change the background color consistently without hacking a lot on the custom Css. 

**Problem/Solution**

Using a variable which defines the background color. That is also possible with Tailwind.

**Alternatives Considered**

Simply changing the background of the `body`, but some elements have their own background and the colors do not match then.

Visual-Dawg:
Overall, I think it would be nice, if the theme uses variables for its colors which fall back to the Tailwind colors, if not defined

Visual-Dawg:
I found this Tailwind plugin https://github.com/mertasan/tailwindcss-variables ,which should make this much easier. 
Turning everything into a variable would be too much work I think, but background and text colors would be nice :)

Visual-Dawg:
what would also be cool is if the Tailwind config gets exposed or if it can be overridden somehow

Visual-Dawg:
`PostCss` is used by default by Tailwind, so plugins work with it by default.

And yes, exposing the config and thus requiring the whole JS techstack would defeat the purpose of using Hugo :D

Visual-Dawg:
Some variables which would be nice:
- `max-width` for the whole thing (with `normal`, `wide` and `full` being preset values)
- Text color palette (a Tailwind palette)
- Background color palette (a Tailwind palette)
- Primary color palette (a Tailwind palette)

This should cover 99% of all theming things I think

I am not sure about a nice way to let the user provide Tailwind palettes though. 
The most flexible would be that each variable is one color, like `fg-500`, `bg-950` and `primary-300`. 
This with a reference to tools like https://www.tints.dev/purple/A855F7 and https://uicolors.app/create would make it easy to customize and provide those palettes based on one color.

Visual-Dawg:
Some variables which would be nice:
- `max-width` for the whole thing (with `normal`, `wide` and `full` being preset values)
- Text color palette (a Tailwind palette)
- Background color palette (a Tailwind palette)
- Primary color palette (a Tailwind palette)

This should cover 99% of all theming things I think

I am not sure about a nice way to let the user provide Tailwind palettes though. 
The most flexible would be that each variable is one color, like `fg-500`, `bg-950` and `primary-300`. 
This with a reference to tools like https://www.tints.dev/purple/A855F7 and https://uicolors.app/create would make it easy to customize and provide those palettes based on one color.

Visual-Dawg:
[...]

PR #632, opened 2025-03-26
chore: migrate to Tailwind CSS v4.0

imfing:
This PR upgrades the CSS build engine to [Tailwind CSS v4](https://tailwindcss.com/blog/tailwindcss-v4), bringing improved performance, reduced dependency footprint, and support for CSS theme variables.

A key benefit of this upgrade is that theme customization is now easier for Hextra users — instead of overriding utility classes, users can simply adjust CSS variables, see https://github.com/imfing/hextra/issues/292 for more context.

As side effect, the upgrade brings **BREAKING CHANGES**, notably the utitlity classes now prefix with `hx:` instead of `hx-`.


kowyo:
Hi @imfing, besides replacing `hx-` with `hx:`, do you have any advice when migrating to Tailwind CSS v4.0?

PR #658, opened 2025-04-04
fix: phantom scroll when `params.footer.enable` is set to `false`

Zabriskije:
When `params.footer.enable` is set to `false`, it causes a phantom scroll effect (more visible on pages with little content). Here's a preview:

footerfalse

The problem is `--menu-height: 3.75rem`, but just setting it to `--menu-height: 3.8rem` fixes the issue.


imfing:
let me look into this, this might actually be a small bug

imfing:
turns out the sidebar bottom menu container was calculated to 61 pixels in the browser, while 3.75rem is 60 pixels. so the total height of sidebar exceeds the view height by 1 pixel

imfing (approved):
thanks for spotting this.

PR #662, opened 2025-04-09
feat: add options for theme toggle and language switching to the navbar

### LINKED ISSUE #654: Allows customization of i18n and theme switch buttons positions
**Feature Description**

Allow customization of the button position for i18n and theme switching to the top (left of the Github icon).

**Problem/Solution**

Currently only the default i18n and theme switch buttons are supported and cannot be customized. Isn't that a good thing if can offer customized locations of them？

Image

Image

**Alternatives Considered**

None currently.

jackieyyang:
Fix #654 
- add `displayThemeToggle` and `displayLanguageSwitch` parameters in `params.navbar`
- when `displayThemeToggle` is true, hide footer and sidebar theme toggle and only show navbar toggle
- when `displayLanguageSwitch` is true, hide footer and sidebar language switch button and only show navbar language switch button
- language switch icon will change to translate when showing in navbar


jackieyyang:
The tailwind dynamic rendering doesn't work every time.
image

Use a different solution solves this problem.
image

The final effect is as follows.
image

imfing:
> Hi, this PR is really useful. What’s the status? Do you think we can get it merged before the v0.10 release?

Hi, I do plan to implement this, but I’d like to take a different approach — componentizing the theme toggle and language switch (kinda similar to #534) instead of controlling their positions via fixed flags. This will require a bit more work.

besides, language switches on the navbar can also be handled via navbar dropdown menu (implemented in https://github.com/imfing/hextra/pull/695)

jackieyyang:
Looking forward to your implement, sound good for implementing them in component.

PR #671, opened 2025-04-21
Add page param to block google indexing 

MattDodsonEnglish:
This param completely blocks a page or directory from being indexed on Google. Useful for avoiding indexing duplicate content or any other page that shouldn't appear in search results.

- Adjust `head.html` for optional `noindex` param, make default value `false`
- Document use in `configuration.md`. I also modified the Flexsearch section to be explicit about what  "search index" means on different parts of the page.

### DIFF
--- a/exampleSite/content/docs/guide/configuration.md
+++ b/exampleSite/content/docs/guide/configuration.md
@@ -243,7 +243,7 @@ There are three available options: `full`, `wide`, and `normal`. By default, the
 
 Similarly, the width of the navbar and footer can be customized by the `params.navbar.width` and `params.footer.width` parameters.
 
-### Search Index
+### FlexSearch Index
 
 Full-text search powered by [FlexSearch](https://github.com/nextapps-de/flexsearch) is enabled by default.
 To customize the search index, set the `params.search.flexsearch.index` parameter in the config file:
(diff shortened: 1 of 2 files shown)


MattDodsonEnglish:
By the way,  I tested the mod on my fork, and the logic is currently used aon https://docs.rhize.com : 
https://github.com/libremfg/rhize-docs/blob/main/layouts/partials/head.html#L4,

At time of writing,  though, only draft  pages at `content/versions` have `noindex`. So all published pages on site have `index, follow` (i.e. the default works). It would be bad to make a mistake.

exampleSite/content/docs/guide/configuration.md
```diff
@@ -302,3 +302,19 @@ services:
...
     ID: G-MEASUREMENT_ID
 ```
+
+### Google Search Index
+
+To [block Google Search](https://developers.google.com/search/docs/crawling-indexing/block-indexing) from indexing a page, set `noindex` to true in your page frontmatter:
+
+```yaml
+title: Configuration (archive version)
+noindex: true
```
imfing, line 312: Let's put it under `params`?

https://gohugo.io/methods/page/param/

imfing (approved):
looks good

PR #674, opened 2025-04-26
feat: allow authors to set the value of reversePagination in page front matter

KStocky:
Currently, there is no way for authors to choose the order in which `PAGE.PrevInSection` and `PAGE.NextInSection` are ordered for navigation in the footer of a page.

This PR just changes the `reversePagination` input in `layouts/partials/components/pager.html` from being hardcoded to `true`, to being set from the page front matter like so:

```yaml
---
params:
  reversePagination: false
---
```

This allows authors to disable this reverse authoring on a per page basis or use the `cascade` front matter to set it on entire page sections.


imfing (commented):
thank you, this looks good

imfing (approved):

PR #681, opened 2025-05-24
Example site: switch to new template system (Hugo v0.146.0)

deining:
This PR adapts the example site to the [new template system in Hugo](https://gohugo.io/templates/new-templatesystem-overview/), introduced with Hugo v0.146.0.


imfing (approved):
@deining cool thanks for the updates

Issue #684, opened 2025-05-26
Publish RSS feed for 'docs' content


quack1:
**Feature Description**

Using this theme, the only RSS feed (index.xml) is created based on the blog posts, and not the 'docs' content. Based on the way we use this theme, it might be useful to be able to configure the theme to choose the directory from which to create the feed.

**Problem/Solution**

Adding a configuration option on hugo.yaml could allow the user to define a list, containing, for each entry : 

* the name of the feed ("Blog", "Doc", etc.), to display it on the front page
* the URL of the feed ("index.xml", "blog.xml", etc.)
* the content directory ("docs/", "blog", etc.)

**Additional Context**

I managed to test that it might be doable by changing line 16 of `layouts/_default/list.rss.xml` to generate a feed from my `docs/` directory : 

```
diff --git a/layouts/_default/list.rss.xml b/layouts/_default/list.rss.xml
index 91db54b..8fde632 100644
--- a/layouts/_default/list.rss.xml
+++ b/layouts/_default/list.rss.xml
@@ -13,7 +13,7 @@
          {{ printf "<atom:link href=%q rel=\"self\" type=%q />" .Permalink .MediaType | safeHTML }}
     {{ end }}
     {{ if not $.Section }}
-      {{ $sections := .Site.Params.rss.sections | default (slice "blog") }}
+      {{ $sections := .Site.Params.rss.sections | default (slice "docs") }}
       {{ .Scratch.Set "rssPages" (first 50 (where $.Site.RegularPages "Type" "in" $sections )) }}
     {{ else }}
       {{ if $.Parent.IsHome }}
```

PR #685, opened 2025-05-26
feat: Add a config value to define the directory to use to generate RSS feed (related to #684)

quack1:
This is related to #684. Does not fixes it, but allows to workaround


imfing:
Isn’t the `.Site.Params.rss.sections` a param used to configure the RSS sections?

PR #691, opened 2025-05-30
fix: CSS loading logic for production and theme dev environments

imfing:
- Updated the CSS loading logic to differentiate between production and theme environments.


jmooring:
Please re-open. The documentation is still wrong.

<https://imfing.github.io/hextra/docs/getting-started/#setup-hextra-as-git-submodule>

It doesn't work.
It is a dead end for new users.

Once you think you have fixed it the documentation, have someone test it starting out in an empty directory.

PR #698, opened 2025-06-01
feat(tags): improve usability of tags

yuri1969:
* Tags can be shown also at Docs (based on 8c4fae2)
* Documented tag-related config flags
* Added example tags to the site
* Made rendered tags active


imfing (approved):
@yuri1969 great work, thanks!

for future: also need to add i18n translations for `tags` & update docs in other languages - can be addressed in a separate PR

PR #723, opened 2025-07-21
fix: giscus theme will be switched to dark when no 'color-theme' field

TwoAnts:


TwoAnts:
When I opened the page for the first time, there was no "color-theme" field in the local-storage of both my PC and mobile browsers. As a result, if I set the default theme for comments to "xxx_light", the comment theme would be replaced with "dark" because getHugoTheme() is not "light".
Maybe you can reproduce this issue by deleting the browser's local storage and refreshing the webpage.

TwoAnts:
In short, the issue encountered is that when the page is opened for the first time, the page may display the light theme, while the gisicus theme is dark.

juzhiyuan:
> In short, the issue encountered is that when the page is opened for the first time, the page may display the light theme, while the giscus theme is dark.

Encountered the same issue just now

imfing (approved):

PR #748, opened 2025-08-14
fix(page-width): add back page width override via css variable

imfing:


imfing:
> I think you can also change:

good point 👍

PR #761, opened 2025-08-16
fix(search): search input and results outline

ldez:
The outline changes to white when the search input is focused.
When using the keyboard, the same weird white outline is applied.

When mixing keyboard and mouse, the outline becomes crazy.

<details>
<summary>before</summary>

https://github.com/user-attachments/assets/bcb96066-ad8e-4da7-94d4-804543fcda71

</details>

<details>
<summary>after</summary>

https://github.com/user-attachments/assets/3380fdac-890b-40a9-96a3-bf9dbb1e4d21

</details>


assets/css/components/search.css
```diff
@@ -14,6 +14,10 @@
     }
   }
 
+  .hextra-search-input:focus {
+    @apply hx:hextra-focus
+  }
```
imfing, line 19: should we add the `hx:focus:hextra-focus` inline to the search input directly?

ldez: yes, as it's a utility, it can be better.

imfing (approved):
looks great thanks!

PR #768, opened 2025-08-19
fix(theme-toggle): quick flash (light/dark)

ldez:
https://github.com/imfing/hextra/pull/766#issuecomment-3202548065

To reproduce the bug:
1. Set dark mode
2. Refresh the page
3. There is a flash

You can also reproduce it in dark mode by switching quickly between pages.


imfing (approved):
looks good

PR #774, opened 2025-08-21
feat(badges): colors and border

ldez:
Allows disabling the border.

Unlike the callouts and GH alert, badges don't need to have "level" type.
So I changed the parameter `type` to `color` to set the color/type of the badge.

I created a compatibility layer, so the previous behavior is still available (`type` parameter and previous type names).

I also added more colors: `green`, `indigo`, `orange`, `amber`, `purple` (the existing colors related to `types` were `gray`, `blue`, `yellow`, `red`).
Those colors were already used by other components (callout, GH alerts).

More colors could be added in the future if it's something you are interested in.


ldez:
I forgot to update the i18n pages, I will add them

PR #778, opened 2025-08-23
feat: support hiding the main sidebar in desktop site

KStocky:
For single pages, the main sidebar is disabled by default, but the screenspace is still being taken up by the empty sidebar. This PR adds a new front matter parameter to allow people to remove it entirely to free up more space on pages that don't need the sidebar


imfing:
> Another possibility is to use the experimental `.context.Params.sidebar.hide`:
> 
> ```diff
> $ git diff layouts/_partials/sidebar.html
> diff --git i/layouts/_partials/sidebar.html w/layouts/_partials/sidebar.html
> index 540261d..637bed5 100644
> --- i/layouts/_partials/sidebar.html
> +++ w/layouts/_partials/sidebar.html
> @@ -12,6 +12,7 @@
>  {{- if .context.Params.sidebar.hide -}}
>    {{- $disableSidebar = true -}}
>    {{- $displayPlaceholder = true -}}
> +  {{- $sidebarClass = "hx:md:hidden hx:xl:hidden" -}}
>  {{- end -}}
> ```

maybe we should just reuse that `hide` option I guess

KStocky:
Hi both!

Regarding modifying the behaviour of `.Params.sidebar.hide`, I didn't want to break any prior behaviour. Hence, the new front matter. But if we are ok with doing that since it is technically marked as experimental, I would not be opposed to it at all.

Regarding a new layout, there is technically the "wide" layout used by the showcase page, but it removes both the main and right sidebar. For blog posts on my site, I still want the right sidebar with the ToC, but not the main left sidebar. My two cents would be that it makes it easier for people to configure their sections if they can enable and disable elements like this in front matter without having to use another layout entirely.

> FYI, removing the aside breaks the title position (it must be centred on the screen) if used inside the blog homepage on a large screen.

Yes, this does seem to be the case, that it breaks the centring on `blog/list.html`
Removing this will make the page behave correctly when the main sidebar is disabled
```html
<div class="hx:max-xl:hidden hx:h-0 hx:w-64 hx:shrink-0"></div>
```
However, when the main sidebar's placeholder is being displayed, the title is off-centre on larger screens..., so I am not sure what the fix would be for that. It at least works well when centring on the page is not required, like with single blog posts.

> You can fix the problem like that: instead of skipping the aside, you could do that:

This is a much-appreciated improvement! Thanks :)

ldez:
As the `hide` was an experimental option (and not used or documented), I think you can change the behavior "safely".
IMHO, this is expected for an experimental option to not be stable.

KStocky:
> As the `hide` was an experimental option (and not used or documented), I think you can change the behavior "safely". IMHO, this is expected for an experimental option to not be stable.

Yeah, fair point. Just updated the PR to now repurpose the `sidebar.hide` parameter to mean both disabling the sidebar and not displaying the placeholder, which has the same effect as the original `disableAll` from the original commit.

If we wanted people to be able to keep the placeholder for layout reasons, we could add a new front matter parameter such as `sidebar.keepPlaceholder` or something like that.

exampleSite/content/docs/guide/configuration.md
```diff
@@ -157,6 +157,21 @@ menu:
...
+
+```yaml {filename="content/docs/guide/configuration.md"}
+---
+title: Configuration
+sidebar:
+  hide: true
+---
+```
+
+This will remove the main sidebar from the page, freeing up space for the main content of the page.
```
imfing, line 172: shall we update the wording to `Hiding the Sidebar` rather than `Disabling`?

imfing (commented):
small comment

imfing (dismissed):

imfing (commented):
After a closer look, I noticed that hiding the sidebar unintentionally breaks the mobile navigation.

This happens because the current implementation couples the sidebar and mobile navigation, which isn’t ideal.
To address this, we should scope the hidden classes to specific elements instead of applying it at the top level.

KStocky:
> After a closer look, I noticed that hiding the sidebar unintentionally breaks the mobile navigation.
> 
> This happens because the current implementation couples the sidebar and mobile navigation, which isn’t ideal. To address this, we should scope the hidden classes to specific elements instead of applying it at the top level.

Did a little playing around with this and yes this PR did break mobile navigation but I think it was because I used the incorrect hidden class. I used "hx:hidden" when I should have used "hx:md:hidden". I have tested this on my android phone now and it seems to be fixed

PR #780, opened 2025-08-24
feat(links): add an option for external link decoration

ldez:
Basically, the same idea as #659, but:
- Disabled by default, with a new option to enable it
- Icons based on Heroicons v1 and inside `data/icons.yaml`
- Uses the partial to add the icons
- Handles RTL and LTR languages
- Uses Tailwind classes

Demo: https://deploy-preview-780--hugo-hextra.netlify.app/blog/v0.10/#technical-improvements

Closes #659


exampleSite/content/docs/guide/configuration.md
```diff
@@ -442,3 +442,12 @@ The field `message` supports Markdown syntax.
...
 If you want to use template syntax, you can define the partial in `layouts/_partials/custom/banner.html`.
 In this case, the field `message` will be ignored.
+
+### Flag Outbound Links
+
+Adds an arrow icon to external links (default: false) when rendering links from Markdown.
+
+```yaml
+params:
+  flagOutboundLinks: true
```
imfing, line 452:
```suggestion
  externalLinkIcon: true
```

layouts/_markup/render-link.html
```diff
@@ -10,7 +10,17 @@
...
-  <a href="{{ $dest | safeURL }}" {{ with .Title }}title="{{ . }}"{{ end }}{{ if strings.HasPrefix .Destination "http" }}target="_blank" rel="noopener"{{ end }}>{{ .Text | safeHTML }}</a>
+{{- $isExternal := strings.HasPrefix .Destination "http" -}}
+<a href="{{ $dest | safeURL }}"
+   {{- with .Title -}}title="{{ . }}"{{- end -}}
+   {{- if $isExternal -}}target="_blank" rel="noopener"{{- end -}}
+>
+  {{- .Text | safeHTML -}}
+  {{- if and .Page.Site.Params.flagOutboundLinks $isExternal -}}
+    {{- partial "utils/icon.html" (dict "name" "arrow-up-right" "attributes" `class="hx:inline hx:rtl:hidden hx:align-baseline" height="1em"`) -}}
+    {{- partial "utils/icon.html" (dict "name" "arrow-up-left" "attributes" `class="hx:inline hx:ltr:hidden hx:align-baseline" height="1em"`) -}}
```
imfing, line 23: Icons here feel slightly heavy
I suggest just using `<span class="external-link-icon"> ↗</span>`. For RTL, we can use css to mirror it to up left.

What do you think?

ldez: The Unicode or symbol character is not really usable because it depends on the system.
In some cases, it can be rendered like that:

<details>



</details>

exampleSite/content/docs/guide/configuration.md
ldez: `externalLinkDecoration`?

layouts/_markup/render-link.html
imfing: Makes sense, we can stick to SVG icons. I believe it would be better to use a single icon per link and leverage CSS for rotation/flip.

ldez: Done

imfing (approved):
looks great!

PR #809, opened 2025-08-30
 feat: new extended cards

### LINKED ISSUE #797: proposal: cards with buttons
**Feature Description**

Adds buttons to cards.

The design of the cards allows adding elements at the bottom.

**Problem/Solution**

Currently, a card is one link, but sometimes, it's useful to have more than one link.

This can be useful, for example, with the showcase.
It allows adding links to the site and to the repository.

**Alternatives Considered**

Creates custom cards.

**Additional Context**

This required considering several challenges (parameters, shortcodes/partials, etc.).

So this is just a raw idea.

I hardcoded the examples.

<details>
<summary>2 buttons with text and icons</summary>

Image
</details>

<details>
<summary4 buttons with only icons</summary>

Image

</details>

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/2fa72276-ed32-485e-8ed6-a9c5b6101d0f

</details>

ldez:
The PR creates new shortcodes to create extended cards.
The extended cards are cards, but users can customize the placement of the title, the image, add buttons, etc.
It allows users to create any shortcode they want to extend the cards.

Each commit is a small step, so you can follow the changes.

Technically, with my approach, the shortcodes are just "interfaces" for partials: they are just here to link the user configuration to the partials.
This design is more flexible and closer to a component design.

The existing card shortcodes and partials are untouched and work as before.

---

The doc is missing because we need to agree together about the approach before writing it.

For now, I just created a draft page to showcase the new extended cards:
- http://localhost:1313/docs/guide/shortcodes/experimental/
- https://deploy-preview-809--hugo-hextra.netlify.app/docs/guide/shortcodes/experimental/

I used the prefix `experimental/`, but this must be changed.
I have some proposals for the prefix: `extended/`, `custom/`, `advanced/`, `hx/`. 

Maybe the CSS is a bit messy, but the render is right (maybe some padding needs to be adjusted)
Note: I mainly copied existing CSS classes from partials into the CSS file.

Fixes #797


imfing:
Hi, thanks for this. On first glance it looks good! I’ll take a closer look at the implementation details later, and organize the CSS part.

> Technically, with my approach, the shortcodes are just "interfaces" for partials: they are just here to link the user configuration to the partials. 

I agree, this is the ideal approach. In fact, we can refactor the previous card to be composed with these partials.

> This design is more flexible and closer to a component design.

Exactly. This feels like a solid example. Nicely done!

imfing (commented):
Thanks! Just a small suggestion to the button styling, the PR implementation looks good overall.

I can imagine some users might mix other shortcodes or Markdown into the new extended card component, so in the future it might be worth adding some note to the docs clarifying how different cases are handled.

IndiTheo:
If it allows to select where to place the image relative to the text, it's great!

IndiTheo:
Hello, are there any expectations for this feature?

@ldez @imfing

ldez:
I need to work on it to rebase it, modify the prefix, and add documentation.

PR #817, opened 2025-09-02
fix(analytics): Matomo analytics

ldez:
I forgot the `with` and to fill the TODO.

### DIFF
--- a/layouts/_partials/components/analytics/matomo.html
+++ b/layouts/_partials/components/analytics/matomo.html
@@ -3,6 +3,16 @@ Matomo Analytics.
 https://developer.matomo.org/guides/tracking-javascript-guide
 */ -}}
 
+{{- with .Site.Params.analytics.matomo -}}
+
+{{- if not .serverURL }}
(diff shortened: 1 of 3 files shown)


layouts/_partials/components/analytics/matomo.html
```diff
@@ -3,6 +3,16 @@
 https://developer.matomo.org/guides/tracking-javascript-guide
 */ -}}
 
+{{- with .Site.Params.analytics.matomo -}}
```
imfing, line 6: this is also checked at https://github.com/imfing/hextra/blob/cc5884dd2a1fa27fe94cd2a750c07a3020683f4a/layouts/_partials/components/analytics/analytics.html#L15

I guess we don't need both?

ldez: :thinking: If someone wants to override (to a specific analytics), he can either override `layouts/_partials/components/analytics/analytics.html` or add an element inside `layouts/_partials/custom/head-end.html`.

So I think I will use `with`.

imfing (approved):

PR #824, opened 2025-09-07
Hextra shortcodes not in sidebar / fixed

jmetzger:
That hextra-shortcodes are not in the sidebar. 
This is annoying, when you are working on a project, because you can only find it through the search.


imfing:
Thanks for the PR! 

This was actually intentional, these are internal Hextra shortcodes, and I avoided putting all shortcodes in the sidebar since they’ll likely be grouped into a dedicated section later.

jnonino (approved):

PR #834, opened 2025-09-13
Enhanced llms.txt support

davlgd:
This PR adds some features to llms.txt support, to better follow [proposal spec](https://llmstxt.org/#example):
- Link to markdown files
- Add an Optional section (use it with `llmsTxtOptional: true` in content front matter)

I only link 1st level sections, ordered alphabetically, but feel free to use different rules


imfing:
I believe it’s possible to disable llmstxt on a per-page basis using:
https://gohugo.io/configuration/outputs/#outputs-per-page

Would it also make sense to add documentation for the new Markdown output format?

hugo.toml
```diff
@@ -8,3 +8,6 @@
     baseName = 'llms'
     mediaType = 'text/plain'
     isPlainText = true
+  [outputFormats.markdown]
```
imfing, line 11: is it necessary here as hugo has `markdown` output format builtin?

https://gohugo.io/configuration/output-formats/

layouts/llms.txt
```diff
@@ -2,35 +2,49 @@
...
+{{- range $pages }}
+- [{{ .Title }}]({{ .Permalink }}index.html.md): {{ .Summary | plainify | truncate 100 | strings.TrimSpace }}{{ if .Date }} - Published {{ .Date.Format "2006-01-02" }}{{ end }}
+{{- end }}
+{{- end }}
+{{- end }}
 
-## Root Pages
-{{- range $rootPages }}
-- [{{ .Title }}]({{ .Permalink }}): {{ .Summary | plainify | truncate 100 | strings.TrimSpace }}{{ if .Date }} - Published {{ .Date.Format "2006-01-02" }}{{ end }}
+## Optional
```
imfing, line 25: this should be omitted if there's no optional pages / sections.

layouts/llms.txt
```diff
@@ -2,35 +2,49 @@
 
 > {{ .Site.Params.description }}
 
-{{ range $section := site.Sections }}
-{{- template "llms-section-tree" dict "context" . "level" 2 }}
-{{ end }}
+{{- if and (not .Params.llmsTxtOptional) (gt (len .Pages) 0) }}
+{{- $sortedSections := sort .Site.Sections "Title" }}
```
imfing, line 6: does the spec require to sort sections by title?

imfing (commented):
have a few questions. in addtion, given that we are adding param for "optional" content, would it make sense to also support exclude a page / section from `llms.txt`?

for example:

```yaml
params:
  llmstxt:
    optional: true # default `false`
    exclude: true  # default `false`
```

davlgd:
> have a few questions. in addtion, given that we are adding param for "optional" content, would it make sense to also support exclude a page / section from `llms.txt`?
> 
> for example:
> 
> ```yaml
> params:
>   llmstxt:
>     optional: true # default `false`
>     exclude: true  # default `false`
> ```

Yep, for example you maybe want to exclude blog from such listing to focus on core documentation (it's why there is an optional section at the beginning, to list useful but not mandatory documents). 

I've also added an exclude parameter working on a per page/section basis (you can exclude a full section or only a single page).

PR #840, opened 2025-09-16
feat: add hoverable dropdown effect on menubar

kowyo:
Added hover interactions: dropdowns now open on mouse enter and close on mouse leave, providing a more intuitive navigation experience.

https://github.com/user-attachments/assets/981e230f-e498-45ab-9aac-cb34d894b7b5


ghac101:
Is there any parameter to customize if one would want this feature?

Issue #714, opened 2025-06-30
Search returns fewer than expected results


bloovis:
I have implemented search for my site as documented in the Hextra configuration guide, but I get fewer results than expected for searches.

My blog has 202 posts, and 20 of them contain the word "Firefox".  But when I search for "firefox", only two of the posts are displayed.  I cannot discern a pattern to this unexpected behavior.

Another strange thing about search is that I can find one of the missing posts by searching for "firefox scrollbar", because that phrase appears in the post.  But "firefox" alone will not return that post.

Here are the relevant lines from hugo.yaml:

```yaml
params:
  search:
    enable: true
    type: flexsearch
    flexsearch:
      index: content
```

bloovis:
After some console.log debugging, I see that the problem is related to the hardcoded 5 in these two places in flexsearch.js:

```js
    const pageResults = window.pageIndex.search(query, 5, { enrich: true, suggest: true })[0]?.result || [];
...
      // Show the top 5 results for each page
      const sectionResults = window.sectionIndex.search(query, 5, { enrich: true, suggest: true, tag: { 'pageId': `page_${result.id}` } })[0]?.result || [];
```

imfing:
> After some console.log debugging, I see that the problem is related to the hardcoded 5 in these two places in flexsearch.js:
> 
> ```js
>     const pageResults = window.pageIndex.search(query, 5, { enrich: true, suggest: true })[0]?.result || [];
> ...
>       // Show the top 5 results for each page
>       const sectionResults = window.sectionIndex.search(query, 5, { enrich: true, suggest: true, tag: { 'pageId': `page_${result.id}` } })[0]?.result || [];
> ```
> 
> The other problem is that `sectionResults` is zero for most pages, so those pages never get put into the results.

thanks, will look into this

imfing:
maybe we could make the number of results configurable? here's a quick diff by Claude:

```diff
diff --git a/assets/js/flexsearch.js b/assets/js/flexsearch.js
index 33c107b..14e8596 100644
--- a/assets/js/flexsearch.js
+++ b/assets/js/flexsearch.js
@@ -318,7 +318,11 @@ document.addEventListener("DOMContentLoaded", function () {
     }
     resultsElement.classList.remove('hx:hidden');
 
-    const pageResults = window.pageIndex.search(query, 5, { enrich: true, suggest: true })[0]?.result || [];
+    // Configurable search limits with sensible defaults
[...]

PR #847, opened 2025-09-29
fix: work around flexsearch returning fewer than expected result pages

bloovis:
This is my fix for the problem described in [issue 714](https://github.com/imfing/hextra/issues/714).  I am a Javascript newbie (though I've been working in assembly language and C since 1976), so there are likely better ways to do things.

I also see that my commit comment doesn't follow the expected norms, which I was not aware of until 10 seconds ago.


assets/js/flexsearch.js
```diff
@@ -327,12 +330,13 @@ document.addEventListener("DOMContentLoaded", function () {
       const result = pageResults[i];
       pageTitleMatches[i] = 0;
 
-      // Show the top 5 results for each page
-      const sectionResults = window.sectionIndex.search(query, 5, { enrich: true, suggest: true, tag: { 'pageId': `page_${result.id}` } })[0]?.result || [];
+      const sectionResults = window.sectionIndex.search(query,
```
imfing, line 333: do we need to pass the limit here as the second argument?

yuri1969:
This PR also works for me - it resolves the annoying issue that caused the majority of search hits to be missing.

@imfing Can we get this reviewed, please?

assets/js/flexsearch.js
imfing: got it, I'll do a quick look, and merge it after

imfing (approved):

Issue #849, opened 2025-10-01
Adding a hero icon to heading 2 and 3 causes on-page menu (right menu) to show jibberish


thedevilfish:
**Description**

Adding hero icons using instructions provided in: https://imfing.github.io/hextra/docs/guide/shortcodes/icon/ causes righthand sidebar to display `HAHAHUGOSHORTCODE11s1HBHB`

**Steps To Reproduce**

1. Started hugo on localhost with the following parameters: `hugo server --buildDrafts --disableFastRender`

2. Create a page with the following markdown

```md
# {{<icon "ticket">}} Ticket Icon in title

This is a ticket section

## {{<icon "exclamation">}} Exclamation Icon in title

This is an exclamation section

### {{<icon "arrow-right">}} Arrow Right Icon in title

This is an arrow right section

## {{<icon "arrow-left">}} Arrow Left Icon in title

This is an arrow left section
```

3. Point to localhost:1313 and see the results

**Expected Behavior**

Icons should render or be excluded from the navigation rendering

**Actual Behavior**

righthand sidebar to display `HAHAHUGOSHORTCODE11s1HBHB`

**Screenshots**

See result:
<img width="1006" height="642" alt="Image" src="https://github.com/user-attachments/assets/d016c1c5-a9fd-4eb2-8af2-60e9115e3c28" />

**Environment**

- Hugo Version: `hugo v0.150.0+extended+withdeploy darwin/arm64 BuildDate=2025-09-08T13:01:12Z VendorInfo=brew`
- Browser/OS: `Chrome / MacOS`
- Theme Version: `0.11.1`

**Additional Context**

N/A

PR #854, opened 2025-10-07
Stripping `HAHAHUGOSHORTCODExxxx` in toc.html when icons are used in headings

thedevilfish:
This is a very quick fix to remove icon shortcodes from rendering as `HAHAHUGOSHORTCODExxxx` on the table of contents on the right hand side of the page. This PR should temporarily strip this until a better solution to actually render the icon properly is devised.

### DIFF
--- a/layouts/_partials/toc.html
+++ b/layouts/_partials/toc.html
@@ -79,7 +79,7 @@
     {{- if .Title }}
       <li class="hx:my-2 hx:scroll-my-6 hx:scroll-py-6">
         <a class="{{ $class }} hx:inline-block hx:text-gray-500 hx:hover:text-gray-900 hx:dark:text-gray-400 hx:dark:hover:text-gray-300 hx:contrast-more:text-gray-900 hx:contrast-more:underline hx:contrast-more:dark:text-gray-50 hx:w-full hx:break-words" href="#{{ anchorize .ID }}">
-          {{- .Title | safeHTML | plainify | htmlUnescape }}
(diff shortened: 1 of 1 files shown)


thedevilfish:
This should fix Issue #849

thedevilfish:
@imfing kindly review and provide any quick feedback.

layouts/_partials/toc.html
```diff
@@ -79,7 +79,7 @@
     {{- if .Title }}
       <li class="hx:my-2 hx:scroll-my-6 hx:scroll-py-6">
         <a class="{{ $class }} hx:inline-block hx:text-gray-500 hx:hover:text-gray-900 hx:dark:text-gray-400 hx:dark:hover:text-gray-300 hx:contrast-more:text-gray-900 hx:contrast-more:underline hx:contrast-more:dark:text-gray-50 hx:w-full hx:break-words" href="#{{ anchorize .ID }}">
-          {{- .Title | safeHTML | plainify | htmlUnescape }}
+          {{- replaceRE "HAHAHUGOSHORTCODE[0-9a-zA-Z]+HBHB" "" (.Title | safeHTML | plainify | htmlUnescape) }}gi
```
imfing, line 82: why there is a `gi` at the end?

imfing (commented):
IMO, this approach relies on internal Hugo heuristics and string patterns that aren’t part of the public API. If Hugo changes its implementation or the shortcode prefix slightly, this workaround could easily break

The recommended way is to avoid including shortcodes (or any HTML) directly in headings in the first place

PR #820, opened 2025-09-03
feat(codeblock): add full URL support

wharris23:
### Summary
This PR adds support for using a `full_url` when linking files in code blocks.

### Changes
- If `full_url` is provided, it is used directly.
- Otherwise falls back to `base_url + filename` or just `filename`.

This keeps the old behavior while allowing more flexibility for cases where file URLs don’t follow the standard format.

Closes #801

Issue #801, opened 2025-08-27
Support full URL when linking to file within code block


imfing:
### Discussed in https://github.com/imfing/hextra/discussions/799

<div type='discussions-op-text'>

<sup>Originally posted by **Aerosand** August 27, 2025</sup>
https://imfing.github.io/hextra/versions/latest/docs/guide/syntax-highlighting/#link-to-file

The way to link a file in a code block is not convenient now when I write my own docs.

> The base_url attribute provides a base URL that will be combined with the file name to generate a link.

Some file url may not follow this kind of format.

Is it more general to use the file's full url directly?</div>

wharris23:
For this issue I am proposing to replace this line:

```go
{{- $file_url := urls.JoinPath $base_url $filename -}}
```

with the following logic that supports an optional full_url, while also keeping the old logic for base_url + filename:

```go
{{- $file_url := "" -}}
{{- if $full_url -}}
  {{- $file_url = $full_url -}}
{{- else if $base_url -}}
  {{- $file_url = urls.JoinPath $base_url $filename -}}
{{- else -}}
  {{- $file_url = $filename -}}
{{- end -}}
```

Declaring $file_url first ensures the variable exists before the conditional assignments.

If full_url is provided, it will be used directly. Otherwise, it falls back to combining base_url and filename, or just using filename if no base_url exists. This way the old method still works while also supporting full URLs that do not fit the old standard formatting.

imfing:
This looks reasonable. Could you please submit a pull request?

PR #882, opened 2025-11-23
feat: Override link text for code blocks

KStocky:
This is a proposed fix for https://github.com/imfing/hextra/issues/801.

I have made this PR because https://github.com/imfing/hextra/pull/820 seems to have been abandoned.

There is a difference in implementation with my PR, though.

Instead of providing another way to specify the URL, this PR proposes to allow writers to override the link text.

## Why implement it this way?

By providing a new option called `full_url` you are giving people another unrelated way of creating URLs that is not compatible with the current way. This may lead to unexpected behaviour (e.g. what should happen if a `base_url` and `full_url` are specified?). Additionally, since it would be adding a new way of creating URLs, the logic to implement it is messier.

With the `name_override` way, we are not introducing a new way of specifying URLs. We are instead just allowing people to override the text of the generated URL with another string.

With this method the URL will ALWAYS be generated with `base_url` + `filename`.
With the change being the link text is `named_override | default filename`

I feel like this is a bit cleaner both in terms of expectations from the user, and from an implementation perspective.


imfing:
Thanks for the explanation! The reason I originally preferred the `full_url` approach is that it allows authors to *completely* override the generated URL instead of always using `base_url` as a mandatory prefix.

With the current model, I’m not sure how one could easily support cases where the final URL shouldn't include the `base_url` at all.

KStocky:
To be clear, with this approach, `base_url` can be used as the full URL. `base_url` + `filename` still provides the full URL. Except we now allow `filename` to be empty.

I guess that is the negative to this approach. `base_url` implies that it can't be the full URL.

PR #904, opened 2026-01-19
fix: respect `search.enable` in the mobile

illia-v:
The sidebar displays the search box regardless of the `search.enable` setting.

When `search.enable` is false, the search box is still displayed, but the related scripts are not loaded:
https://github.com/imfing/hextra/blob/a9ef7ad57dcf2a04d7e85691c58b840b9e52163f/layouts/_partials/scripts/search.html#L2-L26


imfing (approved):
thank you for the fix

PR #905, opened 2026-01-19
fix: respect the page parameter `toc` in the mobile dropdown

illia-v:
`toc: false` can be set for a page to disable its table of contents, the parameter is not respected in the sidebar though.

This is how a page with `toc: false` appears in a sidebar.
image


imfing (approved):
thank you for the fix

PR #914, opened 2026-01-28
fix(glossary): do not list Glossary by default

yuri1969:
The Glossary page used to be shown even when not used at all.


imfing:
this feels a bit hacky. Would it make sense to move `content/glossary.*.md` to the theme docs site, since we shouldn't inject content into users' site?

we could update the documentation to instruct users to create `content/glossary.*.md` within their own site

PR #933, opened 2026-02-19
fix: Site.Author -> Site.Params.Author for Hugo v1.156.0

Fenyutanchan:
Hugo v1.156.0 release removed some deprecated methods, especially `.Site.Author`.
I change it to `.Site.Params.Author` according to the release note of https://github.com/gohugoio/hugo/releases/tag/v0.156.0


MattDodsonEnglish:
This one might be nice to get into a release. Version v0.11.1  breaks with the latest hugo, v0.156.0.

People who don't need RSS, though, can bypass the issue by disabling rss in the top-level of the config.

```yaml
disableKinds: "rss"
```

Issue #937, opened 2026-02-22
Images missing in multilingual Page Bundles due to Markdown render hook


MioYiSama:
### Bug description
When using Hugo's multilingual feature (i18n) with Page Bundles (e.g., having `_index.md`, `_index.zh-cn.md`, and `image 1.png` in the same directory), images render perfectly in the default language but return a 404 error in the translated pages. 

The translated page attempts to fetch the image from the localized path (e.g., `/zh-cn/.../image.png`) instead of resolving to the correct physical resource path generated by Hugo's default language fallback.

### Root cause
The current `layouts/_default/_markup/render-image.html` handles absolute paths (`/`) and individual sibling pages (`$isPage`). However, it misses the logic to properly fetch local resources within a **Page Bundle** using `.Page.Resources.GetMatch`. 

Additionally, if an image filename contains spaces (e.g., `image 1.png`), Markdown automatically encodes it as `%20`, which causes Hugo's `GetMatch` to fail to find the resource.

### Proposed Solution
I have improved the `render-image.html` hook. The updated code now checks `.Page.Resources.GetMatch` to properly resolve the `.RelPermalink` of the bundle resource. This automatically generates the correct absolute path regardless of the current language. 

It also includes two robustness improvements:
1. Replaces `%20` back to spaces using Hugo's native `replace` function to prevent match failures.
2. Uses wildcard matching (`*path*`) to seamlessly handle relative path styles like `./image.png`.

Here is the updated `render-image.html` code:

```go
{{- $alt := .PlainText | safeHTML -}}
{{- $lazyLoading := .Page.Site.Params.enableImageLazyLoading | default true -}}
{{- $enableImageZoom := .Page.Site.Params.imageZoom.enable | default false -}}
{{- if not (eq .Page.Params.imageZoom nil) -}}
  {{- $enableImageZoom = .Page.Params.imageZoom -}}
{{- end -}}
{{- $dest := .Destination -}}
{{- $url := urls.Parse $dest -}}

{{- $isLocal := not $url.Scheme -}}
{{- $isPage := and (eq .Page.Kind "page") (not .Page.BundleType) -}}
{{- $startsWithSlash := hasPrefix $dest "/" -}}
{{- $startsWithRelative := hasPrefix $dest "../" -}}

{{- if and $dest $isLocal -}}
  {{- if $startsWithSlash -}}
    {{- with or (.PageInner.Resources.Get $url.Path) (resources.Get $url.Path) -}}
[...]

PR #938, opened 2026-02-22
fix(render-image): use page resources for relative images in i18n page bundles

MioYiSama:
Related issue: #937


imfing (approved):
Thank you for the pull request, I simplified it a bit. Will release it in the next patch version soon

PR #981, opened 2026-04-22
fix: remove inline TOC click handler so default Hextra can avoid unsafe-inline (CSP)

jecc1982:
## Summary
This PR removes the inline `onClick` handler from the TOC “Scroll to top” button and wires the behaviour through the existing bundled JavaScript instead.

This is a minimal CSP-focused fix that helps vanilla Hextra deployments avoid requiring `unsafe-inline` for this particular interaction.

## What Changed
- Removed the inline `onClick="scrollUp();"` from the TOC back-to-top button
- Reused the existing `assets/js/core/back-to-top.js` implementation
- Attached the click behavior with `addEventListener() `on page load

reference: https://content-security-policy.com/unsafe-hashes/ 

## Why
Inline event handlers require `unsafe-inline` under `script-src`, which makes CSP harder to lock down.

With this change, the default TOC back-to-top behavior no longer depends on inline script.

## Scope
This PR does not fully eliminate all unsafe-inline requirements across Hextra.

There are still other components that use inline scripts today, including examples such as:

- Google Analytics
- Mermaid
- MathJax
- Giscus
- Asciinema
- Medium Zoom
- Matomo
 
## Testing
Verified the TOC back-to-top button still works

### DIFF
--- a/assets/js/core/back-to-top.js
+++ b/assets/js/core/back-to-top.js
@@ -3,6 +3,7 @@
 document.addEventListener("DOMContentLoaded", function () {
   const backToTop = document.querySelector("#backToTop");
   if (backToTop) {
+    backToTop.addEventListener("click", scrollUp);
     document.addEventListener("scroll", (e) => {
       if (window.scrollY > 300) {
         backToTop.classList.remove("hx:opacity-0");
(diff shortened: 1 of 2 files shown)


imfing (approved):

imfing:
Thank you for the PR @jecc1982

PR #989, opened 2026-04-30
fix: resolve page-relative URLs in details shortcode

muit:
Hello,

I noticed my images were not showing up inside details blocks, and needed an extra "../" in the path compared to normal markdown.

This should fix it and keep relative paths to the .md file itself.
Of course you may think of a better solution :)

Awesome theme, Thank you!
Muit

### DIFF
--- a/layouts/_shortcodes/details.html
+++ b/layouts/_shortcodes/details.html
@@ -15,6 +15,6 @@ A built-in component to display a collapsible content.
     <strong class="hx:text-lg">{{ $title | markdownify }}</strong>
   </summary>
   <div class="hx:p-2 hx:overflow-hidden">
-    {{ .InnerDeindent | markdownify }}
+    {{ .InnerDeindent | $.Page.RenderString (dict "display" "block") }}
   </div>
 </details>


imfing:
This change is needed because `details` body content is full page Markdown, not just an inline snippet. Hugo now documents `RenderString` as the right API when Markdown render hooks need the current `.Page` context; `markdownify` can render Markdown, but Hugo specifically recommends `RenderString` when render hooks access `.Page` context. That matters here because Hextra image/link render hooks need the page context to resolve page-bundle and page-relative URLs correctly inside the shortcode.

Using `$.Page.RenderString (dict "display" "block")` is also intentional: Hugo documents `display: "block"` as preserving paragraph wrappers, while `markdownify` removes the wrapper for single-paragraph output. So the expected side effect is that simple text inside `details` may become `<p>...</p>`, but the benefit is that inner Markdown is rendered consistently with normal page content.

References:
- https://gohugo.io/functions/transform/markdownify/
- https://gohugo.io/methods/page/renderstring/

imfing (approved):
Thanks, this is the right fix for page-relative resources inside the details shortcode. The switch to Page.RenderString gives render hooks the current page context, and the paragraph-wrapper behavior from block rendering is expected.

PR #1015, opened 2026-07-12
fix(search): handle inline markup in heading fragments

### LINKED ISSUE #1012: Search index: link-headings break section content extraction, cascade to empty/mis-attributed content across the page
**Description**

On my site (https://aimlbling-about.ninerealmlabs.com/blog/, https://github.com/ahgraber/AIMLbling-about), hextra's built-in search frequently fails to return a complete list of relevant posts, even when words are plainly in the body text; and, when it does return a match, the match is often attributed to the wrong heading/anchor. 

With the help of an AI agent, I was able to trace this to a bug around heading parsing when the heading is a markdown link:

`layouts/_partials/utils/fragments.html` splits page content into per-heading fragments for `search-data.json` by matching on the heading's plain-text title. Headings that are themselves markdown links (e.g. `### [Some Title](https://example.com)`) render with the title wrapped in an `<a>` tag, so the split token never matches. A single failed split doesn't just drop that section's content — it silently swallows all content processed so far in the bottom-up loop, so every heading _above_ the link-heading in the document also ends up empty.

This is exacerbated by an issue (already fixed on `main` but not released) where `assets/js/flexsearch.js` parses section keys (`id#title`) by splitting naively on `#`, so any heading title that contains a `#` (e.g., "the #1 thing") gets truncated in the displayed search result.

On pages where most or all headings are link-headings (my "AI Treadmill" link-roundup posts, and reference sections in regular posts), the fragments bug wipes out the section-level index almost entirely.

**Steps To Reproduce**

see https://github.com/ahgraber/AIMLbling-about, recently fixed in #96caa6e

1. Given a post like

   ```markdown
   ## Plain Heading
   Some paragraph text under a plain heading.

   ## [Link Heading](https://example.com)
   Some paragraph text under a link heading.
   ```
2. When build, the `public/en.search-data.json` will have incorrect heading attribution and/or missing content

**Expected Behavior**

Both sections should be indexed with their own content: `Plain Heading` → "Some paragraph text under a plain heading.", `Link Heading` → "Some paragraph text under a link heading." Search should find text under either heading and attribute results to the correct heading/anchor.

**Actual Behavior**

The `Plain Heading` section's content is empty (or, with a third heading present, contains only content that lexically follows it), and/or the paragraph under `Link Heading` is missing or reattached to an unrelated section. With just two headings as above, `Plain Heading`'s content gets swallowed into whatever section is processed after the failed split.

**Screenshots**

**Environment**

- Hextra Version: v0.12.0 (as a Hugo module)
- Hugo Version: v0.155.3-extended
- Config: `search.flexsearch.index: content`

**Additional Context**

imfing:
## Summary

- split search fragments using stable rendered-heading identifiers instead of plain-text heading HTML
- prevent an unmatched heading from consuming content belonging to earlier sections
- preserve content before the first heading in the page-level fragment
- add regression coverage for linked/formatted headings, hash characters, duplicate titles, and preamble content

## Testing

- npm test
- npm run build
- git diff --check

Fixes #1012


imfing:
@copilot resolve the merge conflicts in this pull request

PR #1020, opened 2026-08-07
feat(search): Drop HTML tags from search result

### LINKED ISSUE #1019: Drop HTML tags from search result
**Feature Description**

Drop HTML tags from search result

**Problem/Solution**

When using HTML in titles they show up in the search result with the HTML elements. Instead of trying to render this stripping the tags would make the search result look better.

**Alternatives Considered**

- Not use HTML in titles
- Accept HTML tags in search results
- Properly render HTML in search result
- Support it, but make it configurable/opt in

**Additional Context**

Thanks for this amazing theme!

I noticed that this test already asserts emphasized code but I suspect this isn't necessarily an explicitly desired behavior but rather a result of testing that we do find inline markup in headings. 

https://github.com/imfing/hextra/blob/5d76ba81a3350c0fd8e4193ddcfbbc398eb59382/tests/search-fragments.spec.ts?plain=1#L64

Given this title and the search string `bombsimon`

```sh
#### `title` referencing <b>bombsimon</b>
```

Current:

Image

With `plainify` and `htmlUnescape` applied to the title:

Image

Relates to https://github.com/golangci/golangci-lint/discussions/6626

bombsimon:
Fixes #1019

As written in the issue, one alternative is to make this an opt-in if it's consider a breaking change or non sensible default. Since I suspect it isn't necessarily a deliberate feature to show the tags in the search result initially I opted to always drop them.

Let me know if you want me to put it behind config (default off) or just close out the PR (and issue) if this isn't wanted.

### DIFF
--- a/layouts/_partials/utils/extract-headings.html
+++ b/layouts/_partials/utils/extract-headings.html
@@ -21,7 +21,8 @@ The scratchpad must be initialized with empty slices before calling this functio
     {{- $.scratch.Add "endTags" (slice "") -}}
     {{- $.scratch.Add "ids" (slice "") -}}
   {{- else -}}
-    {{- $key := (printf "%s#%s" $heading.ID $heading.Title) -}}
+    {{- $headingTitle := $heading.Title | plainify | htmlUnescape | strings.TrimSpace -}}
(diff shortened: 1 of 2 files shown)


imfing (approved):
Thank you, stripping tags is a sensible default

PR #1025, opened 2026-08-20
feat: add option to pass directory to site.Params.math.katex.assets

Keonik1:
# Problem
In `params.math.katex.assets`, you have to specify a whole list of files, which is inconvenient if there are many of them (such as fonts), because it bloats the `hugo.yaml` file

# Solution
Added the ability to specify a directory so that all files in it are included, without having to list every file individually (combined approaches also work)

## Before
```yaml
# hugo.yaml
params:
  math:
    engine: katex
    katex:
      css: “css/katex.min.css”
      assets:
        - css/fonts/KaTeX_AMS-Regular.ttf
        - css/fonts/KaTeX_AMS-Regular.woff
        - css/fonts/KaTeX_AMS-Regular.woff2
        - css/fonts/KaTeX_Caligraphic-Bold.ttf
        ...
```

## Now
```yaml
# hugo.yaml
params:
  math:
    engine: katex
    katex:
      css: “css/katex.min.css”
      assets:
        - css/fonts/
```

### DIFF
--- a/layouts/_partials/scripts/katex.html
+++ b/layouts/_partials/scripts/katex.html
@@ -73,19 +73,30 @@
   {{- end -}}
 {{- end -}}
 
-{{- /* Optionally publish files (fonts, css, js, etc.) from assets and emit tags for css/js with integrity and crossorigin */ -}}
+{{- /* Optionally publish files (fonts, css, js, etc.) or directories (ends with /) from assets and emit tags for css/js with integrity and crossorigin */ -}}
 {{- with site.Params.math.katex.assets -}}
   {{- range . -}}
-    {{- with resources.Get . -}}
-      {{- $name := .Name | lower -}}
-      {{- if strings.HasSuffix $name ".css" -}}
-        {{- $built := . | fingerprint "sha512" -}}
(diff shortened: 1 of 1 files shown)


imfing:
Thanks @Keonik1, this is a nice quality-of-life improvement.

A few comments:

1. **Docs**: please update the `assets` example in `docs/content/docs/guide/configuration.md` (the "Local and Mirrored Script Assets" section) to show the directory form and mention the trailing-slash convention. That section currently carries the exact warning this PR solves. English docs are enough; also note there that any `.css`/`.js` files found in a matched directory will be *loaded* as tags, not just published.
2. **Warn on zero matches**: entries that resolve to nothing are silently skipped — e.g. `css/fonts` without the trailing slash, or a typo'd glob. Could you add something like `{{ warnf "KaTeX asset %q matched no resources" . }}` when an entry matches nothing? We use `warnf` this way elsewhere in the theme.
3. Nit: `{{ $files := "" }}` → `{{ $files := slice }}` would read better.

Also, since the glob branch (`*`) is part of this change, let's document it alongside the directory form.

PR #1030, opened 2026-09-04
fix(mermaid): correct diagram scaling with reduced motion

### LINKED ISSUE #944: Mermaid diagrams not filling the space they reserve
**Description**

After upgrading to Hextra 0.12, mermaid diagrams are rendered incorrectly. The rendered diagram only fills a fraction of the space it reserves. The issue is observable on the [Hextra documentation page](https://imfing.github.io/hextra/docs/guide/diagrams/) as well.

**Steps To Reproduce**

1. Open https://imfing.github.io/hextra/docs/guide/diagrams/ and look at the first graph
2. Open the old version of the same documentaiton page at https://imfing.github.io/hextra/versions/v0.11/docs/guide/diagrams/ and look at the first graph

**Expected Behavior**

Mermaid diagrams should fill the space they reserve on the page (vertically)

**Actual Behavior**
Mermaid diagrams are displayed with lots of whitespace on the bottom and right. This only seems to be true in some cases though, since the sequence diagram in the Hextra documentation looks the same as before.

**Screenshots**

Hextra v0.12 documentation:

Image

Hextra v0.11. documentation:
Image

**Environment**

- Hugo Version: [e.g., 0.156.0]
- Browser/OS: [Firefox and Chrome on MacOS]
- Theme Version: [e.g., v0.12.0]

yuri1969:
Closes #944

It seems I've finally found the root cause of the elusive #944. The reproducibility issue was caused by OS/browser animation configuration.

One of my systems was configured to use no animations which was interpreted by browser and broke the SVG scaling. The others were not affected.

It would be great to verify this fix actually works for other users - check the [mermaid section of the netlify preview](https://deploy-preview-1030--hugo-hextra.netlify.app/docs/guide/diagrams/#mermaid).

---

Systems with animations off triggered the reduced motion style path which affected sizing calculations during Mermaid label measurement.

The `transition-property` affected also SVG geometry.


imfing (approved):
Thanks @yuri1969 for tracking this down.  The cause largely makes sense to me.

The `prefers-reduced-motion` block with `transition-duration: 0.01ms !important` was introduced in #924, which first shipped in v0.12.0, aligning with the regression window reported in #944.

